PR #54, opened 2024-05-02
RED-90: fix some small rotation issues

PudgyPug:
fix the upper bound check for rotation and also turn on rotation edges for 10 node networks


mhanson-github:
Closing for forseeable future.

PR #93, opened 2024-05-28
RED-17: (Prev:Shm 4266) check and implement isOrig that should be standardized in all gossip handlers + test 

dnlbui:
[Enhanced Gossip Endpoint Validations](https://linear.app/shm/issue/RED-17/check-and-implement-isorig-that-should-be-standardized-in-all-gossip)

Standardize gossip endpoint handling with key checks and validations:

Quarter Check: Limits operations to the first two quarters.
Payload Integrity: return and warn if payload, sender, and tracker are null.
Type Validation: Validates payload and signature against schemas.
Signer Verification: Checks signer's presence and matches sender to the original signer.
Quarter Restrictions: Processes original requests only in the 1st quarter.
Message Processing: Updates state and determines message forwarding.


ahmxdiqbal (dismissed):

src/p2p/Join/routes.ts
```diff
@@ -484,6 +540,8 @@ const gossipUnjoinRequests: P2P.P2PTypes.GossipHandler<UnjoinRequest, P2P.NodeLi
   sender: P2P.NodeListTypes.Node['id'],
   tracker: string
 ) => {
+  // unjoin requests should be received at all quarters
```
ahmxdiqbal: this is no longer going to be true after RED-13 is merged. go ahead and the same quarter checks to this handler as well. we will just have to ensure that this is merged after RED-13

ahmxdiqbal:
We should merge RED-13 before this

mhanson-github:
This is merged to dev already.

PR #100, opened 2024-06-03
RED-111: Send 200 response from unjoin on success.  Do not send unjoin request if node not in standby

urnotsam:


ahmxdiqbal (dismissed):

afostr (approved):
This looks great

PR #125, opened 2024-06-17
fix: current-cycle-hash resp type issue

tanuj-shardeum:


mhanson-github:
This is already in dev.

PR #126, opened 2024-06-17
RED-114: use getTxs functions for cycle tx records

urnotsam:


src/p2p/Join/index.ts
```diff
@@ -566,7 +604,7 @@ export function parseRecord(record: P2P.CycleCreatorTypes.CycleRecord): P2P.Cycl
         if (record.finishedSyncing.includes(nodeId)) continue
 
         // add node to lostAfterSelection to be added to the cycle record next cycle
-        lostAfterSelection.push(nodeId)
+        lostAfterSelection.push({ nodeId })
```
ahmxdiqbal: im curious about the point of this change

src/p2p/Join/v2/syncStarted.ts
```diff
@@ -82,7 +79,7 @@ export function addSyncStarted(syncStarted: StartedSyncingRequest): SyncStartedR
     }
   }
 
-  insertSyncStarted(syncStarted.nodeId)
+  newSyncStarted
```
ahmxdiqbal: shouldn't we be setting a pair in the map here?

src/p2p/Join/index.ts
```diff
@@ -199,15 +204,13 @@ export function calculateToAccept(): number {
...
-    join: requestsCopy,
-    startedSyncing: [],
-    finishedSyncing: [],
-    standbyRefresh: [],
+    standbyAdd: drainNewJoinRequests(),
+    startedSyncing: drainSyncStarted(),
+    finishedSyncing: drainFinishedSyncingRequest(),
+    standbyRefresh: drainNewStandbyRefreshRequests(),
+    standbyRemove: drainNewUnjoinRequests(),
+    lostAfterSelection: drainLostAfterSelectionNodes(),
```
ahmxdiqbal: we use cycle txs for changes that we use gossip for. lostAfterSelection is calculated, so we shouldn't include it here. better to just use the old way we did it

ahmxdiqbal: I think i explained this a bit more in the PR for the lib types change

src/p2p/Join/index.ts
ahmxdiqbal: looking at the lib-types changes more closely, i get it now. once we go back to the old way of handling lostAfterSelection, ig this will go back to the way it was

src/p2p/Join/index.ts
```diff
@@ -251,17 +259,35 @@ export function updateRecord(txs: P2P.JoinTypes.Txs, record: P2P.CycleCreatorTyp
...
+        /* prettier-ignore */ if(logFlags.important_as_error) warn(`join:updateRecord:standbyAdd: node not found: ${publicKey}`)
+      }
     }
 
     // ... and unjoining nodes to the standbyRemove field ...
-    for (const publicKey of drainNewUnjoinRequests()) {
-      record.standbyRemove.push(publicKey)
+    for (const request of txs.standbyRemove) {
+      const publicKey = request.sign.owner
+      const node = NodeList.byPubKey.get(publicKey)
```
ahmxdiqbal: This line wont work for standbyRemove. standby nodes are not on the Nodelist.byPubKey array, so this will always return false. the map that you want to use is the standbyNodesInfo one or whatever its exact name is

src/p2p/Join/index.ts
```diff
@@ -251,17 +259,35 @@ export function updateRecord(txs: P2P.JoinTypes.Txs, record: P2P.CycleCreatorTyp
 
   if (config.p2p.useJoinProtocolV2) {
     // for join v2, add new standby nodes to the standbyAdd field ...
-    for (const standbyNode of drainNewJoinRequests()) {
-      record.standbyAdd.push(standbyNode)
+    for (const request of txs.standbyAdd) {
+      const publicKey = request.sign.owner
+      const node = NodeList.byPubKey.get(publicKey)
```
ahmxdiqbal: This line also wont work for standbyAdd. these nodes are unknown to any of the nodelists if they arent even on the standby list. so we wont be able to do a check at all

src/p2p/Join/index.ts
```diff
@@ -276,18 +302,30 @@ export function updateRecord(txs: P2P.JoinTypes.Txs, record: P2P.CycleCreatorTyp
...
     // drain our list of standby refresh TXs to the list
     // not sure full network gossip is the way to go here...
     // what about TX sharing in cycle process?
     // the standby nodes would use a tell to get an active node to submit the tx?
     // only the node that the refresh request was posted to would have the tx?
-    for (const standbyRefresh of drainNewStandbyRefreshRequests()) {
-      record.standbyRefresh.push(standbyRefresh.publicKey)
+    for (const request of txs.standbyRefresh) {
+      const publicKey = request.sign.owner
+      const node = NodeList.byPubKey.get(publicKey)
```
ahmxdiqbal: this will not work. we need to check the standby map, not a nodelist

ahmxdiqbal (dismissed):

PR #127, opened 2024-06-17
RED-105: use currentCycleMarker from memory instead of recalculating it for every tx

PudgyPug:


urnotsam (dismissed):

afostr (approved):
There is a a logic problem but I am going to address this in a follow up PR.  This is because I want to stack my changes as something that can be A/B testing otherwise I would adjust here

PR #128, opened 2024-06-17
RED-134: Remove usage of calculateToAccept() in favor of V2

urnotsam:


ahmxdiqbal:
Since the only place it was changed was in a place where joinProtocolV2 is turned off, we're good. I cant imagine we're ever going to turn joinProtocolV2 off

ahmxdiqbal:
closing per my above comment

PR #138, opened 2024-06-20
BLUE-125 - fix: unbounded array memory leak

Glitch18:
We found another potential memory leak in these screenshots with the debugTXHistory array which was not being cleared and was growing without bounds. Added a fix to clear this array regularly

https://shardeum.slack.com/archives/C06S02QHF98/p1718872627608449


Glitch18:
I have found another significant memory leak. We have an array txDebugStatList which holds debug info. This data is only ever accessed by us when we use  endpoints from debug-app. But the array is constantly populated every time an item is removed from transaction queue, which is quite often. Although there is an option to clear this array, it is only through a debug endpoint, and the validator does not clean it up by itself. This results in another unbounded array resulting in a memory leak. In my last 2 heap screen shots, this had grown upto 220-230 MB. I've added a fix such that the node will clear this array regularly.

PR #141, opened 2024-06-21
GOLD-140: apoptosis ajv hardening and unit tests

S0naliThakur:
Linear: https://linear.app/shm/issue/GOLD-140
Summary: apoptosis internal endpoint ajv hardening and unit tests

### DIFF
--- a/src/p2p/Apoptosis.ts
+++ b/src/p2p/Apoptosis.ts
@@ -117,11 +117,6 @@ const apoptosisInternalRoute: P2P.P2PTypes.Route<InternalBinaryHandler<Buffer>> 
 
       const req = deserializeApoptosisProposalReq(requestStream)
 
-      const errors = verifyPayload(AJV_IDENT.APOPTOSIS_REQ, req)
-      if (errors && errors.length > 0) {
-        warn('apoptosisInternalRoute: bad request:', errors)
-        return response(BadRequest('Request validation errors'), serializeResponseError)
-      }
       const apopProposal: P2P.ApoptosisTypes.SignedApoptosisProposal = {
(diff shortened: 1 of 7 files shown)


src/p2p/Apoptosis.ts
```diff
@@ -96,17 +100,28 @@ const failExternalRoute: P2P.P2PTypes.Route<Handler> = {
...
-        return
+      const requestStream = getStreamWithTypeCheck(payload, TypeIdentifierEnum.cApoptosisProposalReq)
+      if (!requestStream) {
+        errorHandler(RequestErrorEnum.InvalidRequest)
+        return response(BadRequest('Invalid apoptosisInternalRoute request stream'), serializeResponseError)
       }
+
       const req = deserializeApoptosisProposalReq(requestStream)
+
+      const errors = verifyPayload(AJV_IDENT.APOPTOSIS_REQ, req)
```
kgmyatthu: @arhamj Thought we reached consensus on this to be in the ser/de functions.

src/types/ApoptosisProposalReq.ts
```diff
@@ -37,6 +34,10 @@ export function deserializeApoptosisProposalReq(stream: VectorBufferStream): Apo
     id,
     when,
   }
+  const errors = verifyPayload(AJV_IDENT.APOPTOSIS_REQ, obj)
```
kgmyatthu, line 37: AJV validation are not consistent as to where it happen

kgmyatthu (changes requested):
Let's get double confirmation in next meeting for where ajv verify payload should happen and stick to the pattern.

PR #147, opened 2024-06-24
GOLD-153 fix: serialization sequence fix

kgmyatthu:
URGENT CRITICAL: This commit fixes the incorrect optional field serialization sequence


afostr (changes requested):
It looks like things were working before, but this change will cause a failure.   This is because 0 is used on the read side to represent data it seems "=== 0" .

This fix could be changes in both places, and that would be ok.  (this can remain part of the larger set of group changes.  will not need and emergency fix on it.

arhamj:
I don't think this is true can you recheck the de-serialiser?

kgmyatthu:
Has this got into the dev?

arhamj:
This is merged

PR #148, opened 2024-06-24
RED-127 gossip join requests without double queueing

ahmxdiqbal:
whenever we received a join request, we used to always save it to be gossiped next cycle, even if we received it before sendRequests(the time to send a joinRequest) was called. This was achieved with a double queue. This forced waiting caused network spin-up to take longer. We have gotten rid of it now in favor of a single queue and handled the issues that come with that approach.

### DIFF
--- a/src/p2p/Join/v2/index.ts
+++ b/src/p2p/Join/v2/index.ts
@@ -249,6 +249,8 @@ export function isOnStandbyList(publicKey: string): boolean {
 }
 
 export function isInNewJoinRequests(publicKey: string): boolean {
+  return newJoinRequests.some((node) => node.nodeInfo.publicKey === publicKey);
+}
   if (newJoinRequests.find((node) => node.nodeInfo.publicKey === publicKey)) {
     return true
   } else {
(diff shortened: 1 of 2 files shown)


src/p2p/Join/v2/index.ts
```diff
@@ -248,6 +248,14 @@ export function isOnStandbyList(publicKey: string): boolean {
   }
 }
 
+export function isInNewJoinRequests(publicKey: string): boolean {
```
urnotsam, line 251: This could be simplified

```suggestion
export function isInNewJoinRequests(publicKey: string): boolean {
  return newJoinRequests.some((node) => node.nodeInfo.publicKey === publicKey);
}
```

src/p2p/Join/v2/index.ts
```diff
@@ -248,6 +248,16 @@ export function isOnStandbyList(publicKey: string): boolean {
...
+  return newJoinRequests.some((node) => node.nodeInfo.publicKey === publicKey);
+}
+  if (newJoinRequests.find((node) => node.nodeInfo.publicKey === publicKey)) {
```
ahmxdiqbal: stupid github. need to fix this before merge, but in the middle of debugging rn

PR #153, opened 2024-06-26
GOLD-156 fix binary_sync_trie_hashs dangerous node identity reference.

kgmyatthu:
In binary_sync_trie_hashes, the identity of the node is reference from the sender_id, which an askBinary invoker can put any type of string including other node's id. In Handler this node (derived from the sender_id) is then passed into a list of voters to kickstart, potentially creating frictions in the network protocol


kgmyatthu:
Closed, shifting the new approach at the abstraction level before reaching the app defined handler.

PR #157, opened 2024-06-28
Blue 134

jairajdev:
- type fix in pruneNetworkChangeQueue dapp function
- updated to remove old entries from appliedConfigChanges that are no longer in the changes list


src/shardus/index.ts
```diff
@@ -2936,10 +2934,12 @@ class Shardus extends EventEmitter {
       const changeHash = this.crypto.hash(change)
       //skip handled changes
       if (this.appliedConfigChanges.has(changeHash)) {
+        activeConfigChanges.add(changeHash)
```
urnotsam, line 2995: the comment says //skip handled changes but now we are adding the hash to activeConfigChanges.  Is this intended?

mhanson-github:
This is already in dev.

PR #160, opened 2024-07-01
BLUE-147 POQo Implementation

Glitch18:
Implements the POQo algorithm


src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1024,220 @@ class TransactionConsenus {
...
+    Comms.registerInternal(
+      'poqo-send-vote',
+      async (
+        payload: AppliedVoteHash,
+        _respond: unknown,
+        _sender: unknown,
+        _tracker: string,
+        msgSize: number
+      ) => {
+        profilerInstance.scopedProfileSectionStart('poqo-send-vote', false, msgSize)
```
arhamj, line 1465: We can add checks to ensure that the receiver is an execution group node and is an elected aggregator.

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1024,220 @@ class TransactionConsenus {
...
+      'poqo-send-receipt',
+      (
+        payload: AppliedReceipt2,
+        _respond: unknown,
+        _sender: unknown,
+        _tracker: string,
+        msgSize: number
+      ) => {
+        profilerInstance.scopedProfileSectionStart('poqo-send-receipt', false, msgSize)
+        try{
```
arhamj, line 1339: Add catch blocks on all endpoints with counters to avoid unexpected crashes.

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1024,220 @@ class TransactionConsenus {
...
+      'poqo-data-and-receipt',
+      async (
+        payload: {
+          finalState: { txid: string; stateList: Shardus.WrappedResponse[] }, 
+          receipt: AppliedReceipt2
+        }, 
+        _respond: unknown,
+        _sender: string,
+      ) => {
+        profilerInstance.scopedProfileSectionStart('poqo-data-and-receipt')
```
arhamj: Verify FACT sender

arhamj: https://linear.app/shm/issue/GOLD-165

arhamj: This is already handled.

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1024,220 @@ class TransactionConsenus {
...
+            //let wrappedResponse = data as Shardus.WrappedResponse
+            //this.queueEntryAddData(queueEntry, data)
+            if (data == null) {
+              /* prettier-ignore */ if (logFlags.error && logFlags.verbose) this.mainLogger.error(`poqo-data-and-receipt data == null`)
+              continue
+            }
+            // validate corresponding tell sender
+            if (_sender == null) {
+              /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`poqo-data-and-receipt invalid sender for data: ${data.accountId}, sender: ${_sender}`)
+              continue
```
kun6fup4nd4: why not break? _sender is not defined in the loop, should be checked outside.

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1024,239 @@ class TransactionConsenus {
...
+        profilerInstance.scopedProfileSectionStart('poqo-send-vote', false, msgSize)
+        try{
+          const queueEntry = this.stateManager.transactionQueue.getQueueEntrySafe(payload.txid)
+          if (queueEntry == null) {
+            /* prettier-ignore */ nestedCountersInstance.countEvent('poqo', 'poqo-send-vote: no queue entry found')
+            return
+          }
+          const collectedVoteHash = payload as AppliedVoteHash
+          // We can reuse the same function for POQo
+          this.tryAppendVoteHash(queueEntry, collectedVoteHash)
```
jairajdev, line 1480: We should also check that sign owner of the voteHash is from the executionGroupNodes. inside the `tryAppendVoteHash` fn.
A malicious executionGroup node can create bad voteHash and sign by using different keys, we should check if all these sign owner are in the exectionGroupNodes and reject them if they are not there.

PR #162, opened 2024-07-02
GOLD-164: Signature majority validation for non locally produced receipts

jintukumardas:
A new method **verifyTwoThirdsMajority** has been added to the **TransactionConsensus** class, which takes receipt and executionGroupNodes as input parameters. This method verifies if the receipt has more than 2/3rd of the valid signers by calculating the required majority.

### DIFF
--- a/src/state-manager/TransactionConsensus.ts
+++ b/src/state-manager/TransactionConsensus.ts
@@ -1044,6 +1044,14 @@ class TransactionConsenus {
           }
 
           if (logFlags.verbose) this.mainLogger.debug(`POQo: received receipt from gossip for ${queueEntry.logID} forwarding gossip`)
+
+          const executionGroupNodes = new Set(queueEntry.executionGroup.map((node) => node.publicKey))
+          const hasTwoThirdsMajority = this.verifyAppliedReceipt(payload, executionGroupNodes)
+          if (!hasTwoThirdsMajority) {
+            /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`Receipt does not have the required majority for txid: ${payload.txid}`)
+            return
+          }
+          
           queueEntry.poqoReceipt = payload
           queueEntry.appliedReceipt2 = payload
           queueEntry.recievedAppliedReceipt2 = payload
@@ -1271,6 +1279,12 @@ class TransactionConsenus {
           }
           if (!queueEntry.hasSentFinalReceipt) {
             if (logFlags.verbose) this.mainLogger.debug(`POQo: received data & receipt for ${queueEntry.logID} starting receipt gossip`)
+            const executionGroupNodes = new Set(queueEntry.executionGroup.map(node => node.publicKey));
+            const hasTwoThirdsMajority = this.verifyAppliedReceipt(payload.receipt, executionGroupNodes)
+            if(!hasTwoThirdsMajority) {
+              /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`Receipt does not have the required majority for txid: ${payload.receipt.txid}`)
+              return
+            }
             queueEntry.poqoReceipt = payload.receipt
             queueEntry.appliedReceipt2 = payload.receipt
             queueEntry.recievedAppliedReceipt2 = payload.receipt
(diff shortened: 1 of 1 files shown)


src/state-manager/TransactionConsensus.ts
```diff
@@ -1220,6 +1238,31 @@ class TransactionConsenus {
     )
   }
 
+  verifyTwoThirdsMajority(receipt: AppliedReceipt2, executionGroupNodes: Set<string>): boolean {
```
arhamj: nit: `verifyAppliedReceipt` sounds like a better name.

src/state-manager/TransactionConsensus.ts
```diff
@@ -1220,6 +1238,31 @@ class TransactionConsenus {
...
+    const vote = receipt.appliedVote; 
+    const voteHash = this.calculateVoteHash(vote);
+    const appliedVoteHash = {
+      txid: vote.txid,
+      voteHash,
+    }
+
+    let validSignatures = 0;    
+    for (const owner of uniqueSignOwners) {
+      const signedObject = { ...appliedVoteHash, sign: receipt.signatures.find(sign => sign.owner === owner) };
```
arhamj: Can we use a map from owner to sign? Since this code is going to be a hot path it will be more efficient

src/state-manager/TransactionConsensus.ts
```diff
@@ -1220,6 +1238,31 @@ class TransactionConsenus {
     )
   }
 
+  verifyTwoThirdsMajority(receipt: AppliedReceipt2, executionGroupNodes: Set<string>): boolean {
+    const uniqueSignOwners = new Set(receipt.signatures.map(sign => sign.owner));
+    const totalNodes = executionGroupNodes.size;
+    const requiredMajority = Math.ceil((2 / 3) * totalNodes);
```
arhamj: Looks like we use a Math.round at consensus level: https://github.com/shardeum/shardus-core/blob/e8e32714bb86884f97aa387fcfd2b9e481cffb0e/src/state-manager/TransactionConsensus.ts#L1717-L1718

src/state-manager/TransactionConsensus.ts
```diff
@@ -1220,6 +1238,31 @@ class TransactionConsenus {
...
+
+    const vote = receipt.appliedVote; 
+    const voteHash = this.calculateVoteHash(vote);
+    const appliedVoteHash = {
+      txid: vote.txid,
+      voteHash,
+    }
+
+    let validSignatures = 0;    
+    for (const owner of uniqueSignOwners) {
```
abdulazeem-tk4vr: this should come from the execution group.

abdulazeem-tk4vr: Having unit tests for this separately would be nice to see how it works.

kgmyatthu:
Has this been tested in a running network?

abdulazeem-tk4vr:
It would be nice to see if we can compare the signatures from both the node lists for verification.

src/state-manager/TransactionConsensus.ts
```diff
@@ -1039,6 +1039,13 @@ class TransactionConsenus {
             // We've already send this receipt, no need to send it again
             return
           }
+
+          const executionGroupNodes = new Set(queueEntry.executionGroup.map((node) => node.id))
+          const hasTwoThirdsMajority = this.verifyAppliedReceipt(payload, executionGroupNodes)
+          if (!hasTwoThirdsMajority) {
+            /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`Receipt does not have the required majority!`)
```
kgmyatthu: Nit: More logging in this block. Tx id?

kgmyatthu:
> It would be nice to see if we can compare the signatures from both the node lists for verification.

What do you mean by "both the node list" ? @abdulazeem-tk4vr

jintukumardas:
> Has this been tested in a running network?

I have tested the changes on the local network, but we plan to deploy them to a larger server with all the POQo modifications for further testing @kgmyatthu

abdulazeem-tk4vr:
> > It would be nice to see if we can compare the signatures from both the node lists for verification.
> 
> What do you mean by "both the node list" ? @abdulazeem-tk4vr

Jintu addressed that in the recent commit. Comparing the signature from the receipt with the nodes from the execution group.

kgmyatthu (approved):

abdulazeem-tk4vr (approved):

PR #169, opened 2024-07-05
GOLD 169 : Binary Migration of PoQo - send receipt endpoint

abdulazeem-tk4vr:
Linear Task :[Send Receipt Migration]( https://linear.app/shm/issue/GOLD-169/poqo-send-receipt-migration-to-binary)

### DIFF
--- a/src/state-manager/TransactionConsensus.ts
+++ b/src/state-manager/TransactionConsensus.ts
@@ -85,6 +85,7 @@ import {
 import { BadRequest, InternalError, NotFound, serializeResponseError } from '../types/ResponseError'
 import { randomUUID } from 'crypto'
 import { Utils } from '@shardus/types'
+import { deserializePoqoSendReceiptReq } from '../types/PoqoSendReceiptReq'
 
 class TransactionConsenus {
   app: Shardus.App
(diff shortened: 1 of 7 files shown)


src/state-manager/TransactionConsensus.ts
```diff
@@ -1213,6 +1214,58 @@ class TransactionConsenus {
...
+          if (queueEntry == null) {
+            /* prettier-ignore */ nestedCountersInstance.countEvent('poqo', 'binary/poqo_send_receipt: no queue entry found')
+            return
+          }
+
+          if (queueEntry.poqoReceipt) {
+            // We've already handled this
+            return
+          }
+
```
jintukumardas, line 1249: payload verification is missing

abdulazeem-tk4vr: verification data needs to be passed and that seems like it is optional unless you can clarify more on what is required.

jintukumardas: Oh, okay. I thought we needed to verify the data in this handler. If it's already verified, then it's fine.

PR #170, opened 2024-07-07
RED-167 feat/add sync info to report

ahmxdiqbal:


src/state-manager/AccountPatcher.ts
```diff
@@ -2597,6 +2597,31 @@ class AccountPatcher {
     })
     //todo what about situation where we do not have enough votes??
     //todo?? more utility / get list of oos radix
+
+    // set recentRuntimeSync to true in the radices that have had recent coverage changes
+    for (const coverageChange of this.stateManager.coverageChangesCopy) {
```
ahmxdiqbal, line 2624: @afostr since we are using a copy of coverageChanges, i think that the recentRuntimeSync will only be set for the cycle that we realize that we need to runtime sync. however, it will not be true for the next cycle even if we are still runtime syncing from last cycle. this might not be an issue if runtime syncing is expected to be quick and not span multiple cycles. but if it is, then could we use the sync trackers themselves to set this flag? as in, we check all the current sync trackers and set the flags for the radixes that they are operating for. this would set the flag for the entire duration of the runtime sync rather than just for the cycle it started syncing. wanted to get your thoughts on this idea

PR #171, opened 2024-07-07
updated config settings to be closer to prod ready

mssabr01:
Set logging to prod setup. Error mode might be too intense for prod and allow easy DOSing
startInFatalsLogMode: true, 
startInErrorLogMode: false,

Prod will be able to safely filter bogon IPs
forceBogonFilteringOn: true

Also making the default mode be production instead of debug


mssabr01:
Smoke passed 
https://devs-work.shardeum.org/job/Testing/job/Smoke/job/smoke-stage1/66/

PR #172, opened 2024-07-08
GOLD-172 Migrating PoQo endpoints to Binary

abdulazeem-tk4vr:
Linear Task: https://linear.app/shm/issue/GOLD-172/migrate-new-poqo-endpoints-to-binary
Branch : poqo-binary

### DIFF
--- a/src/types/PoqoSendReceiptReq.ts
+++ b/src/types/PoqoSendReceiptReq.ts
@@ -27,7 +27,7 @@ export function serializePoqoSendReceiptReq(
 
   serializeAppliedVote(stream, obj.appliedVote)
 
-  if (obj.confirmOrChallenge !== undefined) {
+  if (obj.confirmOrChallenge) {
     stream.writeUInt8(1)
     serializeConfirmOrChallengeMessage(stream, obj.confirmOrChallenge)
   } else {
@@ -55,8 +55,6 @@ export function deserializePoqoSendReceiptReq(stream: VectorBufferStream): PoqoS
   let confirmOrChallenge
   if (stream.readUInt8() === 1) {
     confirmOrChallenge = deserializeConfirmOrChallengeMessage(stream)
-  } else {
-    confirmOrChallenge = undefined
   }
 
   const signaturesLength = stream.readUInt16()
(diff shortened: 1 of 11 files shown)


src/types/PoqoSendReceiptReq.ts
```diff
@@ -0,0 +1,86 @@
...
+import { VectorBufferStream } from '../utils/serialization/VectorBufferStream'
+import { deserializeAppliedVote, serializeAppliedVote } from './AppliedVote'
+import {
+  deserializeConfirmOrChallengeMessage,
+  serializeConfirmOrChallengeMessage,
+} from './ConfirmOrChallengeMessage'
+import { deserializeSign, serializeSign } from './Sign'
+import { TypeIdentifierEnum } from './enum/TypeIdentifierEnum'
+
+export type PoqoSendReceiptReq = AppliedReceipt2
```
kgmyatthu: The ser/de and type def for AppliedReceipt is already done, just use the existing function and avoid redundant code to prevent the same entity be having potentially different behavior.

kgmyatthu (changes requested):

src/types/PoqoSendReceiptReq.ts
abdulazeem-tk4vr: serialisation and deserialisation for appliedReceipt2 is not binary complete in the file that you refer to and the type of serialisation is different as well. I don't want to touch that since it uses json stringify.

abdulazeem-tk4vr: We need enhancements and faster processing here for PoQO hence I wanted to give importance to the binary serialised code.

kgmyatthu: > serialisation and deserialisation for appliedReceipt2 is not binary complete in the file that you refer to and the type of serialisation is different as well. I don't want to touch that since it uses json stringify. We need enhancements and faster processing here for PoQO hence I wanted to give importance to the binary serialised code.

Which I'm 100% aware. But this reasoning does not address my original review which is to avoid the situation two ser/de function for the same object type is implemented.  I see the need for perf fine tuning , we can still do this  while reducing this fallacy by moving serialization works to original appliedReceipt2 ser/de and reference it back to anywhere the appliedReceipt2 involved. 

I've made the changes to address this.

kgmyatthu:
image
Local sanity test, 10 node, 1 tps, 50 account, recipient mode: random, 5 Minute with LT2.

src/types/PoqoSendReceiptReq.ts
abdulazeem-tk4vr: thanks!

PR #186, opened 2024-07-11
Poqo + FACT + Rotation fixes

afostr:


afostr:
Note when it is time to finally merge. I'll do a fast forward merge.

src/types/AppliedReceipt2.ts
```diff
@@ -32,41 +32,41 @@ export function serializeAppliedReceipt2(
...
-  // stream.writeUInt16(obj.signatures.length)
-  // // TODO: Convert to for loop
-  // obj.signatures.forEach((sig) => serializeSign(stream, sig))
-  // stream.writeString(obj.app_data_hash)
-  const stringified = Utils.safeStringify(obj)
-  stream.writeString(stringified)
+  stream.writeString(obj.txid)
+  stream.writeUInt8(obj.result ? 1 : 0)
+  serializeAppliedVote(stream, obj.appliedVote)
+  serializeConfirmOrChallengeMessage(stream, obj.confirmOrChallenge)
```
kun6fup4nd4, line 38: AFAIK it is not used in POQo

src/state-manager/TransactionConsensus.ts
```diff
@@ -302,14 +359,27 @@ class TransactionConsenus {
           }
 
           const readableReq = deserializeGetTxTimestampReq(requestStream)
+          // handle rare race condition where we have seen the txId but not the timestamp
+          if (Context.config.p2p.timestampCacheFix && this.seenTimestampRequests[readableReq.txId] && this.txTimestampCacheByTxId[readableReq.txId] == null) {
+            nestedCountersInstance.countEvent('consensus', 'get_tx_timestamp seen txId but found no timestamp')
+            return respond(tsReceipt, serializeGetTxTimestampResp)
```
kun6fup4nd4, line 365: what the point serializing empty tsReceipt if result is not found? why not have errors like 400, 404. Or binary protocol is not supporting it ?

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+            null,
+            queueEntry.transactionGroup,
+            false,
+            4,
+            payload.txid,
+            '',
+            true
+          )
+          queueEntry.hasSentFinalReceipt = true
+        } finally {
```
kun6fup4nd4, line 1139: no catch

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+      'poqo-receipt-gossip',
+      (payload: AppliedReceipt2) => {
+        profilerInstance.scopedProfileSectionStart('poqo-receipt-gossip')
+        try {
+          const queueEntry = this.stateManager.transactionQueue.getQueueEntrySafe(payload.txid)
+          if (queueEntry == null) {
+            /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`poqo-receipt-gossip no queue entry for ${payload.txid}`)
+            return
+          }
+          if (queueEntry.hasSentFinalReceipt === true) {
```
kun6fup4nd4, line 1108: was this condition have any impact on gossip ? Gossip design assumed only 1 iteration of sending or there can be multiple, so this condition will create possible gaps?

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+          // validate corresponding tell sender
+          if (_sender == null) {
+            /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`poqo-data-and-receipt invalid sender for txid: ${readableReq.finalState.txid}, sender: ${_sender}`)
+            return
+          }
+
+          const isValidFinalDataSender =
+            this.stateManager.transactionQueue.factValidateCorrespondingTellFinalDataSender(
+              queueEntry,
+              _sender
```
kun6fup4nd4, line 1177: not sure that it was fixed, at least I didn't find - _sender in header and actual sender who signed message can be different

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+            )
+          if (isValidFinalDataSender === false) {
+            /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`poqo-data-and-receipt invalid sender ${_sender} for data: ${queueEntry.acceptedTx.txId}`)
+            return
+          }
+
+          if (!queueEntry.hasSentFinalReceipt) {
+            const executionGroupNodes = new Set(queueEntry.executionGroup.map(node => node.publicKey));
```
kun6fup4nd4, line 1185: just curious how "expensive" this procedure. I saw it is used a lot in POQo

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+            const hasTwoThirdsMajority = this.verifyAppliedReceipt(readableReq.receipt, executionGroupNodes)
```
kun6fup4nd4, line 1186: queueEntry is taken based on readableReq.finalState.txid, but verification is done based on readableReq.receipt.txid

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+
+          const isValidFinalDataSender = this.stateManager.transactionQueue.factValidateCorrespondingTellFinalDataSender(queueEntry, _sender)
+          if (isValidFinalDataSender === false) {
+            /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`poqo-data-and-receipt invalid sender ${_sender} for data: ${queueEntry.acceptedTx.txId}`)
+            return
+          }
+
+          if (!queueEntry.hasSentFinalReceipt) {
+            const executionGroupNodes = new Set(queueEntry.executionGroup.map(node => node.publicKey));
+            const hasTwoThirdsMajority = this.verifyAppliedReceipt(payload.receipt, executionGroupNodes)
```
kun6fup4nd4, line 1317: queueEntry is taken based on readableReq.finalState.txid, but verification is done based on readableReq.receipt.txid

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+
+          if (logFlags.verbose)
+            this.mainLogger.debug(
+              `POQo: Received receipt from aggregator for ${queueEntry.logID} starting CT2 for data & receipt`
+            )
+          const receivedReceipt = readableReq as AppliedReceipt2
+          queueEntry.poqoReceipt = receivedReceipt
+          queueEntry.appliedReceipt2 = receivedReceipt
+          queueEntry.recievedAppliedReceipt2 = receivedReceipt
+          queueEntry.hasSentFinalReceipt = true
```
kun6fup4nd4, line 1503: if gossip will fail or corresponding tell, this true value will not represent fact that it was sent

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+              null,
+              null,
+              Array.from(nodesToSendTo),
+              false,
+              4,
+              queueEntry.acceptedTx.txId
+            )
+            nestedCountersInstance.countEvent(`processing`, `forwarded final data to storage nodes`)
+          }
+        } finally {
```
kun6fup4nd4, line 1390: no catch

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+            null,
+            queueEntry.transactionGroup,
+            false,
+            4,
+            payload.txid,
+            '',
+            true
+          )
+          this.stateManager.transactionQueue.factTellCorrespondingNodesFinalData(queueEntry)
+        } finally {
```
kun6fup4nd4, line 1449: no catch

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+          if (!hasTwoThirdsMajority) {
+            /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`Receipt does not have the required majority for txid: ${payload.txid}`)
+            nestedCountersInstance.countEvent('poqo', 'poqo-send-receipt: Rejecting receipt because no majority')
+            return
+          }
+          const receivedReceipt = payload as AppliedReceipt2
+          queueEntry.poqoReceipt = receivedReceipt
+          queueEntry.appliedReceipt2 = receivedReceipt
+          queueEntry.recievedAppliedReceipt2 = receivedReceipt
+          queueEntry.hasSentFinalReceipt = true
```
kun6fup4nd4, line 1435: same here if gossip failed, true value here will not represent that it was sent

src/state-manager/TransactionConsensus.ts
```diff
@@ -1024,6 +1094,565 @@ class TransactionConsenus {
...
+          const collectedVoteHash = payload as AppliedVoteHash
+
+          // Check if vote hash has a sign
+          if (!collectedVoteHash.sign) {
+            /* prettier-ignore */ nestedCountersInstance.countEvent('poqo', 'poqo-send-vote: no sign found')
+            return
+          }
+          // We can reuse the same function for POQo
+          this.tryAppendVoteHash(queueEntry, collectedVoteHash)
+        } finally {
```
kun6fup4nd4, line 1554: no catch

src/state-manager/TransactionQueue.ts
```diff
@@ -308,7 +313,25 @@ class TransactionQueue {
...
           for (const data of payload.stateList) {
-            this.queueEntryAddData(queueEntry, data)
+            if (
+              configContext.stateManager.collectedDataFix &&
+              configContext.stateManager.rejectSharedDataIfCovered
+            ) {
+              const consensusNodes = this.stateManager.transactionQueue.getConsenusGroupForAccount(
+                data.accountId
+              )
+              const coveredByUs = consensusNodes.map((node) => node.id).includes(Self.id)
```
kun6fup4nd4, line 323: is it possible that account will be covered by node, but node will be OOS on this account ?

src/state-manager/TransactionQueue.ts
```diff
@@ -411,15 +435,63 @@ class TransactionQueue {
...
+            const senderNodeId = header.sender_id
+            if (configContext.p2p.useFactCorrespondingTell) {
+              let isSenderOurExeNeighbour = false
+              // check if it is a neighbour exe node sharing data
+              if (configContext.stateManager.shareCompleteData) {
+                const senderIsInExecutionGroup = queueEntry.executionGroupMap.has(senderNodeId)
+                const neighbourNodes = utils.selectNeighbors(queueEntry.executionGroup, queueEntry.ourExGroupIndex, 2) as Shardus.Node[]
+                const neighbourNodeIds = neighbourNodes.map((node) => node.id)
+                isSenderOurExeNeighbour = senderIsInExecutionGroup && neighbourNodeIds.includes(senderNodeId)
+                if (isSenderOurExeNeighbour) {
```
kun6fup4nd4, line 449: all 4 above lines can be done outside of the loop of accounts.

src/state-manager/TransactionQueue.ts
```diff
@@ -411,15 +435,63 @@ class TransactionQueue {
...
+            } else {
+              isSenderValid = this.validateCorrespondingTellSender(queueEntry, state.accountId, senderNodeId)
+            }
+
+            if (this.stateManager.testFailChance(configContext.debug.ignoreDataTellChance, 
+            'ignoreDataTellChance', queueEntry.logID, '', logFlags.verbose ) === true ) {
+              isSenderValid = false
+            }
+
+            if (i !== 0 && isSenderValid === false) {
```
kun6fup4nd4, line 477: what is the logic under i !== 0 condition ? why senderinvalid is only important when it is second iteration?

src/state-manager/TransactionQueue.ts
```diff
@@ -460,6 +535,16 @@ class TransactionQueue {
               /* prettier-ignore */ if (logFlags.error && logFlags.verbose) this.mainLogger.error(`broadcast_finalstate data == null`)
               continue
             }
+            // validate corresponding tell sender
+            if (_sender == null ||  _sender.id == null) {
+              /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`broadcast_finalstate invalid sender for data: ${data.accountId}, sender: ${JSON.stringify(_sender)}`)
+              continue
+            }
+            const isValidFinalDataSender = this.factValidateCorrespondingTellFinalDataSender(queueEntry, _sender.id)
```
kun6fup4nd4, line 543: in broadcast_data we are using configContext.p2p.useFactCorrespondingTell condition, why we are not using here?

src/state-manager/TransactionQueue.ts
```diff
@@ -557,40 +642,16 @@ class TransactionQueue {
               /* prettier-ignore */ if (logFlags.error && logFlags.verbose) this.mainLogger.error(`broadcast_finalstate data == null`)
               continue
             }
+            const isValidFinalDataSender = this.factValidateCorrespondingTellFinalDataSender(queueEntry, header.sender_id)
```
kun6fup4nd4, line 645: same above question here

src/state-manager/TransactionQueue.ts
```diff
@@ -2559,8 +2628,21 @@ class TransactionQueue {
       )
     }
     if (queueEntry.collectedData[data.accountId] != null) {
-      // we have already collected this data
-      return
+      if (configContext.stateManager.collectedDataFix) {
+        // compare the timestamps and keep the newest
+        const existingData = queueEntry.collectedData[data.accountId]
+        if (data.timestamp > existingData.timestamp) {
```
kun6fup4nd4, line 2634: this will not work if coveredByUs is enabled

src/state-manager/TransactionQueue.ts
```diff
@@ -5536,35 +6112,61 @@ class TransactionQueue {
...
+
+              //TODO check for receipt and move to repair state / await final data
+
+              if(this.config.stateManager.awaitingDataCanBailOnReceipt){
+                const receipt = this.stateManager.getReceipt2(queueEntry)
+                if(receipt != null){
+                  //we saw a receipt so we can move to await final data
+
+                  this.updateTxState(queueEntry, 'await final data', 'receipt while waiting for initial data')
+                  continue
```
kun6fup4nd4, line 6139: so this can cause majority of nodes to go to 'await final data' state

src/types/AppliedReceipt2.ts
afostr: I am pulling in some more commits will see if that helps this

afostr: found out that using the receipt data types from the spec (thanks to this comment!).  have a new task to solve the issue

afostr: will be in a separate pr

src/state-manager/TransactionConsensus.ts
afostr: thanks, have created a new task to look this. (fix will be in a new branch)

src/state-manager/TransactionConsensus.ts
afostr: this may be fixed or removed in another branch.  created  GOLD-210 to verify.  Thanks!

src/state-manager/TransactionConsensus.ts
afostr: this could result in some skipped gossip if it happens enough.  I created new counter, and put this in a spreadsheet of error counters we need to monitor in tests or live networks.

afostr: sys-368

src/state-manager/TransactionConsensus.ts
afostr: gold-209

src/state-manager/TransactionConsensus.ts
afostr: it could be expensive but in the latest code (possibly changed since review time) I only see this in a two spots that happen once per TX so I think it is ok.   the execution group fortunately is limited to 128 that helps some, otherwise I would ask for this to be not declarative.

src/state-manager/TransactionConsensus.ts
afostr: this looks to be covered, but I will get confirmation.
I also have another concern that I thought we had to allow data from both unrwrapped and wrapped index values.   if we don't allow that nodes could be blocking final data that is valid

afostr: going to leave unresolved but not a merge blocker

src/state-manager/TransactionConsensus.ts
afostr: thanks I have added error conditions/logs counters for the case where  if(readableReq.finalState.txid != readableReq.receipt.txid) {

(will be in different PR)

src/state-manager/TransactionConsensus.ts
afostr: confirmed non binary endpoint is deleted here:
https://github.com/shardeum/shardus-core/pull/202/files
otherwise I would add the same error handling as mentioned in last note

src/state-manager/TransactionConsensus.ts
afostr: note the non-binary endpoint with this will be removed.

I think in general our systems to not track if gossip out fails.  I know you have mentioned the good point before in meetings that since it is tcp will know if transport fails on us.    I am adding this general topic an   ARCH review task

ARCH-9

src/state-manager/TransactionConsensus.ts
afostr: the non binary code for this one is getting removed.  the binary version has a catch.   (created after this reveiw)

poqo-data-and-receipt

src/state-manager/TransactionConsensus.ts
afostr: will be deleted in another PR.    new binary endpoint has the catch

src/state-manager/TransactionConsensus.ts
afostr: same comments as before.   created ARCH-9  to consider not checking gossip/tell requests at the higher level

src/state-manager/TransactionConsensus.ts
afostr: also nuked in other pr

confirmed binary endpoint has the catch

src/state-manager/TransactionConsensus.ts
afostr: this pattern does happen a lot so it will likely get squashed in some places in a future perf pass

src/state-manager/TransactionQueue.ts
afostr: the old 'broadcast_state'  for non binary is removed in a new PR.
checked binary endpoint:

It is ok for a node to be OOS and still broadcast the state it has.  We have our patcher / and will get some alerting for if there is persistent OOS that is non patchable.    supermajority consensus will be fine even if some nodes are OOS.  this can happen when a node rotates in and we have upgraded the monitor a ton to help us tell the difference between expected vs unexpected patcher behavior

src/state-manager/TransactionQueue.ts
afostr: great catch, I have corrected locally for my next PR  (in the binary handler)

src/state-manager/TransactionQueue.ts
afostr: oh wow.   I have removed this in a local commit.   will also send to blue team for comments.  
I agree that we should also let the check apply for i === 0

src/state-manager/TransactionQueue.ts
afostr: this is a great point,  in this case I am certain that switched over now with no going back,  will get that config removed in a follow up.

src/state-manager/TransactionQueue.ts
afostr: same as above.  I think there is no going back to  the old corresponding tell with POQo

src/state-manager/TransactionQueue.ts
afostr: I am not certain on the correct outcome here. there was another recent fix in this function to make sure that we don't over count the collected data.

Raising to blue team.

src/state-manager/TransactionQueue.ts
afostr: I think this cant impact a majority because, a receipt now requires a supermajority of nodes  (with POQo)
this helps a node that was stuck waiting for data while 2/3rds+ were able to get the data and form a receipt.

afostr: I added a counter so we can track how often this is activated.  going to start tracking a lot of things like this in automation

src/state-manager/TransactionConsensus.ts
thantsintoe: @arhamj will put in a fix for this today.

src/state-manager/TransactionQueue.ts
thantsintoe: I don't remember exactly but i === 0 seems to be from very old interface where the first item is app data hash (or maybe global data?). I am not sure. But it should be 100% changed to `if (isSenderValid === false)` now.

src/state-manager/TransactionQueue.ts
thantsintoe: This should still work fine even if `rejectSharedDataIfCovered: false` (i.e. we allow other nodes to send us the account that we cover), we should check the timestamp of incoming account data against the one we already collected. We should replace the collected data only if the incoming data is newer than the one we have.

src/state-manager/TransactionConsensus.ts
thantsintoe: https://github.com/shardeum/shardus-core/pull/226

PR #199, opened 2024-07-16
Diff of fixes from PR 186  to latest  poqo

afostr:


src/shardus/shardus-types.ts
```diff
@@ -1213,6 +1215,8 @@ export interface ServerConfiguration {
     forwardToLuckyNodes: boolean
     // should the network forward TXs to lucky nodes?  (only for the nonce queue)
     forwardToLuckyNodesNonceQueue: boolean
+    // forward to multiple lucy nodes (requires other configs to forward at all)
```
kun6fup4nd4, line 1218: typo

src/shardus/shardus-types.ts
```diff
@@ -1257,6 +1261,10 @@ export interface ServerConfiguration {
     removeStuckTxsFromQueue2: boolean
     // how long to wait before removing the tx from the queue. Based on time since first vote seen
     stuckTxRemoveTime2: number
+    // whether to remove the tx from the queue if it stuck for X min. must be in consensing state
+    removeStuckTxsFromQueue3: boolean
+    // how long to wait before removing the tx from the queue. must be in consensing state
+    stuckTxRemoveTime3: number
```
kun6fup4nd4, line 1267: IMHO splitting feature flag to 2 parameters can add mess during support at a later stages.
for example, if stuckTxRemoveTime3 > 0 - feature is enabled, otherwise, disabled

src/state-manager/TransactionConsensus.ts
```diff
@@ -1135,12 +1147,38 @@ class TransactionConsenus {
...
+              if (logFlags.verbose)
+                this.mainLogger.debug(`poqo-receipt-gossip: requesting final data for ${queueEntry.logID}`)
+              nestedCountersInstance.countEvent(
+                'request-final-data',
+                'final data timeout, making explicit request'
+              )
+
+              const nodesToAskKeys = payload.signatures?.map((signature) => signature.owner)
+
+              await this.stateManager.transactionQueue.requestFinalData(
```
kun6fup4nd4, line 1169: requestFinalData can return based on number of accounts found in collectedFinalData === payload.appliedVote.account_id.length which is different from queueEntry.hasValidFinalData. Potentially can be infinite loop.

kun6fup4nd4:
finished with review

src/shardus/shardus-types.ts
afostr: fixed in my branch.  I am the typo king

src/shardus/shardus-types.ts
afostr: I like the two flags because operationally we usually think in terms of turning a feature on/off.  if we have to set the correct tuned time here I think that leaves more for error.

src/state-manager/TransactionConsensus.ts
afostr: I have enhanced out counters and added some to our doc for sys-automation monitoring.

I have created a blue task to resolve this.  the resolution will be that we need to check state information before putting something in the collected final data map.  there is no value in putting a wrong state account this loop which can cause logic errors like the one you found above.

Created blue-203 to resolve

afostr:
this PR was used to help us review partial work.  it is now merged in so closing

PR #205, opened 2024-07-17
RED-13 Support cycle tx for unjoin(standbyRemove)

ahmxdiqbal:


src/p2p/Join/index.ts
```diff
@@ -725,11 +742,54 @@ export function sendRequests(): void {
...
+  if (queuedUnjoinReuqests.length > 0) {
+    for (const unjoinRequest of queuedUnjoinReuqests) {
+      console.log('Join:sendRequests: unjoinRequest', unjoinRequest)
+
+      // TO-DO: address this following:
+      // processing it again to make sure it is still valid but
+      // we already processed it in the unjoin route; can we just gossip it?
+      //
+      // there an issue i've thought of: processNewUnjoinRequest automatically 
+      // adds the unjoin request to the node's unjoin map. Since we call it last cycle in the unjoin handler but dont gossip it,
```
ahmxdiqbal: I think i already handled this by removing the unjoin request the cycle that we receive it. line ref is 229 inside src/p2p/Join/routes.ts

mhanson-github:
This is in dev.

PR #207, opened 2024-07-22
Deterministic fact tell and controllable "concatCorrespondingTellUseUnwrapped" flag

thantsintoe:
This PR includes following fixes to improve stuck txs:

1. fact tell functions should use shard data from deterministic txGroupCycle instead of using currentCycleShardData
2. use "concatCorrespondingTellUseUnwrapped" flag whether to concat extra corresponding indices or not. set "false" by default so that it is easier to  debug stuck txs
3. set "filterReceivingNodesForTXData" to false by default to false for safety over optimization
4. add extra logs and counter for effective debugging of stuck txs


afostr:
this was merged to dev

PR #208, opened 2024-07-22
RED-195: Add max queue entry time in report

urnotsam:


dnlbui:
lgtm 

https://linear.app/shm/issue/RED-195/make-monitor-also-output-the-max-q-time-per-node-vs-the-avgd-time-it

gpt summary: Added functionality to track and report maximum transaction time in queue. Implemented getMax method in Statistics class and max methods in Ring and FifoStats classes. Updated Reporter to include maxTxTimeInQueue in heartbeat data.

mhanson-github:
Already in dev

PR #217, opened 2024-07-30
GOLD-159: AJV addition and verification data check: binary_compare_cert

jintukumardas:
This commit adds AJV validation for the CompareCertReq and CompareCertResp types.


arhamj (approved):
LGTM. We can consider using an enum identifier for AJV structs

PR #223, opened 2024-07-31
Beta1.12.1rc0  updates since waternet fixes 

afostr:
created for review purposes


src/p2p/Active.ts
```diff
@@ -97,7 +97,7 @@ let p2pLogger: Logger
 
 let activeRequests: Map<P2P.NodeListTypes.Node['publicKey'], P2P.ActiveTypes.SignedActiveRequest>
 let queuedRequest: P2P.ActiveTypes.ActiveRequest
-let neverGoActive = false
+export let neverGoActive = false
```
mssabr01, line 100: Should go in the debug config in server.ts

src/p2p/Lost.ts
```diff
@@ -551,6 +549,17 @@ export function parseRecord(record: P2P.CycleCreatorTypes.CycleRecord): P2P.Cycl
     }
   }
 
+  // If we see our node in the lostSyncing field of the cycle record, emit a sync timeout event
+  for (const id of record.lostSyncing) {
+    const node = NodeList.nodes.get(id)
+    if (node == null) {
+      error(`Lost syncing node ${id} is not in the network`)
```
mssabr01, line 556: This comment says emit an event if our node is found in the lostSyncing record, but this is emitting an event for every node found

mssabr01: Is the comment inaccurate? This is part of the slashing logic and every node that is part of an emitted event here gets a slashing tx created for it. Perhaps the comment is backwards.

mssabr01 (approved):

src/p2p/Active.ts
afostr: i agree .  created blue 204 to fix in a following PR

src/p2p/Lost.ts
afostr: comments were wrong . I have confirmed and clarified in my latest branch

afostr:
this PR was used to help us review partial work.  it is now merged in so closing

PR #226, opened 2024-08-05
Fix header and sender check in Comms

arhamj:
The header and sender check in the Comms module has been updated to ensure that the signer and sender are the same. This fix addresses an issue where internal routes were being used by nodes not in the network.

### DIFF
--- a/src/p2p/Comms.ts
+++ b/src/p2p/Comms.ts
@@ -633,21 +633,28 @@ export function registerInternalBinary(route: string, handler: InternalBinaryHan
     }
     /* prettier-ignore */ if (logFlags.verbose && logFlags.p2pNonFatal) console.log('header:', header)
     /* prettier-ignore */ if (logFlags.verbose && logFlags.p2pNonFatal) info(`registerInternalBinary: request info: route: ${route} header: ${Utils.safeStringify(header)} sign: ${Utils.safeStringify(sign)}`)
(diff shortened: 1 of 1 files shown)


src/p2p/Comms.ts
```diff
@@ -634,9 +634,14 @@ export function registerInternalBinary(route: string, handler: InternalBinaryHan
     /* prettier-ignore */ if (logFlags.verbose && logFlags.p2pNonFatal) console.log('header:', header)
     /* prettier-ignore */ if (logFlags.verbose && logFlags.p2pNonFatal) info(`registerInternalBinary: request info: route: ${route} header: ${Utils.safeStringify(header)} sign: ${Utils.safeStringify(sign)}`)
     if (
```
thantsintoe: I think it looks correct. My only concern is that I have to spend a few minutes to understand this if condition check. I hope it will be easier to understand if we can simplify it a bit (something like below). Other than that, the change looks good. 

```const isSenderPublicKeyUnknown = !NodeList.byPubKey.has(sign.owner);
const isSenderIdUnknown = !NodeList.nodes.has(header.sender_id);
const isNotInDebugMode = !config.debug.enableTestMode;

if ((isSenderPublicKeyUnknown || isSenderIdUnknown) && isNotInDebugMode) {
  warn('registerInternalBinary: internal routes can only be used by nodes in the network...');
  return;
}```

PR #233, opened 2024-08-09
RED-247: Add null check for `queueEntry` when getting index value in the TransactionQueue processing loop

dnlbui:
https://shardeum.slack.com/archives/C07F5L9TF33/p1723222226341989?thread_ts=1723210967.343049&cid=C07F5L9TF33
Above is the ask

Summary: Add null check for queue entry in TransactionQueue processing loop

- Prevent potential null error causing node to error out
- Log error and skip to next transaction if queue entry is null

### DIFF
--- a/src/state-manager/TransactionQueue.ts
+++ b/src/state-manager/TransactionQueue.ts
@@ -5609,7 +5609,12 @@ class TransactionQueue {
         this.clearDebugAwaitStrings()
 
         // eslint-disable-next-line security/detect-object-injection
-        const queueEntry: QueueEntry = this._transactionQueue[currentIndex]
+        const queueEntry: QueueEntry | undefined = this._transactionQueue[currentIndex]
+        if (!queueEntry) {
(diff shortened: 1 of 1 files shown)


src/state-manager/TransactionQueue.ts
```diff
@@ -5608,7 +5608,12 @@ class TransactionQueue {
         this.clearDebugAwaitStrings()
 
         // eslint-disable-next-line security/detect-object-injection
-        const queueEntry: QueueEntry = this._transactionQueue[currentIndex]
+        const queueEntry: QueueEntry | undefined = this._transactionQueue[currentIndex]
+        if (queueEntry == null) {
```
urnotsam, line 5613: you could just do !queueEntry.  If you stick with null equality check we should use strict equals (===)

dnlbui: Noted, and looked up why so I don't do this again. Pushed the change.

urnotsam (dismissed):

mhanson-github (approved):
Made one slight change per Andrew's review (in person) but otherwise is good.

PR #236, opened 2024-08-12
RED-196 check trie hashes for near node down check

magonjr:


src/p2p/Lost.ts
```diff
@@ -1192,8 +1192,101 @@ async function isDownCheck(node) {
...
-    if (config.p2p.useProxyForDownCheck) {
-      //using the 'apoptosize' route to check if the node is up.
+    if (config.p2p.useNearNodeForDownCheck) {
+      //using the 'get_trie_hashes' route to check if the node is up.
+      let obj = { counter: currentCycle, checker: Self.id, target: node.id, timestamp: shardusGetTime() }
+      let hash = crypto.hash(obj)
+      let hashTrieReq: HashTrieReq = {
+        radixList: ['0'],
+      }
+      let closestNodes = stateManager.getClosestNodes(hash, 5, true)
```
thantsintoe, line 1202: How do we know if these 5 nodes will cover the radix `0` ?

src/p2p/Lost.ts
```diff
@@ -1192,8 +1192,101 @@ async function isDownCheck(node) {
...
+      if (logFlags.verbose)
+        info(
+          `lost check result for nearNodeAsk near node:${node.id} cycle ${currentCycle} is ${utils.stringifyReduce(
+            nearNodeResponse
+          )}`
+        )
+
+      if (nearNodeResponse == null) {
+        /* prettier-ignore */ nestedCountersInstance.countEvent('p2p', 'isDownCheck-down-near-node-fail', 1)
+        return 'down'
```
thantsintoe, line 1242: Does it mean "near node" is down or "target node" is down? If  we cannot get response from near node, isn't that  we report it as "lost" and try another near node?

src/p2p/Lost.ts
```diff
@@ -1192,8 +1192,101 @@ async function isDownCheck(node) {
...
+
+      if (logFlags.verbose)
+        info(
+          `lost check result for nearNodeAsk target node:${node.id} cycle ${currentCycle} is ${utils.stringifyReduce(
+            targetNodeResponse
+          )}`
+        )
+
+      if (targetNodeResponse == null) {
+        /* prettier-ignore */ nestedCountersInstance.countEvent('p2p', 'isDownCheck-down-near-node-fail', 1)
```
thantsintoe, line 1271: this counter should be `isDownCheck-down-target-node-fail`

src/p2p/Lost.ts
```diff
@@ -1192,8 +1192,101 @@ async function isDownCheck(node) {
...
+      }
+
+      const nodeHashesMap = new Map<string, string>()
+      for (let nodeHash of nearNodeResponse.nodeHashes) {
+        nodeHashesMap.set(nodeHash.radix, nodeHash.hash)
+      }
+      //compare the hashes from the near node and the target node
+      for (let nodeHash of targetNodeResponse.nodeHashes) {
+        if (nodeHashesMap.get(nodeHash.radix) !== nodeHash.hash) {
+          /* prettier-ignore */ nestedCountersInstance.countEvent('p2p', 'isDownCheck-down-near-node-fail', 1)
```
thantsintoe, line 1282: this should be a different counter.

src/p2p/Lost.ts
```diff
@@ -1192,8 +1192,101 @@ async function isDownCheck(node) {
...
+        return 'down'
+      }
+
+      const nodeHashesMap = new Map<string, string>()
+      for (let nodeHash of nearNodeResponse.nodeHashes) {
+        nodeHashesMap.set(nodeHash.radix, nodeHash.hash)
+      }
+      //compare the hashes from the near node and the target node
+      for (let nodeHash of targetNodeResponse.nodeHashes) {
+        if (nodeHashesMap.get(nodeHash.radix) !== nodeHash.hash) {
```
thantsintoe, line 1281: What if the target node is up and responsive but has out of sync data? In that case, we should not mark it  as `down` as the node may repair the radix in next cycle.

src/p2p/Lost.ts
```diff
@@ -1192,8 +1192,101 @@ async function isDownCheck(node) {
   /* prettier-ignore */ if (logFlags.lost) info(`Checking internal connection for ${node.id}, cycle: ${currentCycle}`)
 
   try {
-    if (config.p2p.useProxyForDownCheck) {
-      //using the 'apoptosize' route to check if the node is up.
+    if (config.p2p.useNearNodeForDownCheck) {
+      //using the 'get_trie_hashes' route to check if the node is up.
+      let obj = { counter: currentCycle, checker: Self.id, target: node.id, timestamp: shardusGetTime() }
+      let hash = crypto.hash(obj)
```
thantsintoe, line 1198: why is this hash used instead of a radix covered by target node?

thantsintoe (changes requested):
Please let us know how these changes are properly tested.

PR #240, opened 2024-08-13
RED-201: Network Generated Transactions 

urnotsam:


src/p2p/SyncV2/index.ts
```diff
@@ -169,6 +177,14 @@ function syncStandbyNodeList(activeNodes: P2P.SyncTypes.ActiveNode[]): ResultAsy
   })
 }
 
+function syncTxList(activeNodes: P2P.SyncTypes.ActiveNode[]): ResultAsync<{ hash: string; tx: P2P.ServiceQueueTypes.AddNetworkTx }[], Error> {
+  return robustQueryForTxListHash(activeNodes).andThen(({ value, winningNodes }) =>
+    getTxListFromNode(winningNodes[0], value.txListHash).andThen((txList) =>
+      verifyTxList(txList, value.txListHash).map(() => txList)
+    )
+  )
+}
```
ahmxdiqbal, line 186: i think this need to be more robust. Right now, we get the robust txlisthash and query and check just one node for the its txlist and verify that it matches. However, if it doesn't we should retry a few more times. I think retrying 5 times or the length of winningNodes should be sufficient, whichever is smaller

ahmxdiqbal: I say this because if we get unlucky and ask a node that is OOS and has a different txlist, syncing looks like its gonna fail

src/p2p/ServiceQueue.ts
```diff
@@ -0,0 +1,768 @@
...
+      }
+      if (record.txadd.some((entry) => entry.txData.nodeId === node.id && entry.type === 'nodeReward')) {
+        warn(`shutdown condition: active node with id ${node.id} is already in txadd; this should not happen`)
+        continue
+      }
+
+      // get latest entry for node in txList. and if it is init then we inject otherwise continue
+      // first iterate over txlist backwards and get first entry that has public key of node
+      const txListEntry = reversedTxList.find((entry) => entry.tx.txData.publicKey === node.publicKey)
+      if (txListEntry && txListEntry.tx.type !== 'nodeInitReward') {
```
afostr, line 343: Taking this as techdebt.   We should replace with  app.getFinalNodeTXName()  an similar functions so that the final TX  names, shape and creation come from the app

urnotsam: This is being refactored by Phillipp as this shouldn’t be in the shardus layer but rather in the app layer

urnotsam: All the shutdown impl in update record is going to be changed when he pushes that tomorrow

src/shardus/index.ts
```diff
@@ -1742,6 +1763,19 @@ class Shardus extends EventEmitter {
     return latestCycle ? latestCycle.active : 0
   }
 
+  fastIsPicked(numberToPick: number) {
```
afostr, line 1766: did we need to cut and paste this?

afostr: oh, I see it is wrapping and for nodes. A modified name would be nice.     fastIsPickedActiveNodes or similar

src/p2p/CycleCreator.ts
```diff
@@ -739,6 +746,9 @@ function makeCycleRecord(
     archiverListHash: '',
     standbyNodeListHash: '',
     random: config.debug.randomCycleData ? Math.floor(Math.random() * 1000) + 1 : 0,
+    txadd: [],
+    txremove: [],
+    txlisthash: '',
```
ahmxdiqbal, line 751: i think we should be using camel case for cycle record fields. thats the pattern for all other fields in the cycle record

PR #242, opened 2024-08-14
Allow additional properties in CycleRecord AJV

arhamj:
https://linear.app/shm/issue/GOLD-215

Made the check of properties more lenient for cycle record in compare cert AJV check.


jintukumardas (approved):

mhanson-github:
In dev.

PR #244, opened 2024-08-15
SYS-332, SYS-375, SYS-376, SYS-377, SYS-378: Add `validateJoinRequest.test.ts`

muni-corn:
Adds unit tests for testing `validateJoinRequest` and some of its components.

Refactors were required in order to eliminate any circular dependencies that prevented mock functions from working as intended.

To see this work, just run `npx jest "validateJoinRequest"`.

### DIFF
--- a/src/p2p/Join/validate.ts
+++ b/src/p2p/Join/validate.ts
@@ -27,7 +27,8 @@ export function validateJoinRequest(joinRequest: P2P.JoinTypes.JoinRequest): Joi
     validateJoinRequestHost(joinRequest.nodeInfo.externalIp) ||
     verifyUnseen(joinRequest.nodeInfo.publicKey) ||
     verifyNodeUnknown(joinRequest.nodeInfo) ||
-    validateJoinRequestTimestamp(joinRequest.nodeInfo.joinRequestTimestamp)
+    validateJoinRequestTimestamp(joinRequest.nodeInfo.joinRequestTimestamp) ||
+    null
   )
 }
 
--- a/src/p2p/Join/validate.ts
+++ b/src/p2p/Join/validate.ts
@@ -251,7 +251,7 @@ export const ALREADY_SEEN_ERROR = {
  * If the `joinRequest`'s node has not already been seen this cycle, it returns
  * `null`.
  */
-function verifyUnseen(publicKey: hexstring): JoinRequestResponse | null {
+export function verifyUnseen(publicKey: hexstring): JoinRequestResponse | null {
   // Check if this node has already been seen this cycle
   if (getSeen().has(publicKey)) {
     if (logFlags.p2pNonFatal) nestedCountersInstance.countEvent('p2p', `join-skip-seen-pubkey`)
--- a/src/p2p/Join/validate.ts
+++ b/src/p2p/Join/validate.ts
@@ -1,0 +1,296 @@
+import { version } from '../../../package.json'
+import { logFlags } from '../../logger'
+import { hexstring, P2P } from '@shardus/types'
+import * as utils from '../../utils'
+import { isEqualOrNewerVersion } from '../../utils'
+import { config } from '../Context'
+import * as CycleChain from '../CycleChain'
+import * as NodeList from '../NodeList'
+import { isBogonIP, isInvalidIP, isIPv6 } from '../../utils/functions/checkIP'
+import { nestedCountersInstance } from '../../utils/nestedCounters'
+import { Utils } from '@shardus/types'
+import { JoinRequestResponse } from './types'
+import { info, warn } from './logging'
+import { getAllowBogon, getSeen } from './state'
+
+/**
+ * Returns a `JoinRequestResponse` object if the given `joinRequest` is invalid or rejected for any reason.
+ */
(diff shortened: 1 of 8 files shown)


src/p2p/Join/validate.ts
```diff
@@ -0,0 +1,306 @@
...
+    if (logFlags.p2pNonFatal) nestedCountersInstance.countEvent('p2p', `join-skip-seen-pubkey`)
+    if (logFlags.p2pNonFatal) info(ALREADY_SEEN_ERROR.reason)
+    return ALREADY_SEEN_ERROR
+  }
+
+  // Mark node as seen for this cycle
+  getSeen().add(publicKey)
+
+  return null
+}
```
chrypnotoad: The AI review mentioned this one, but it seems wrong. I think the single line ifs confuse it.  
> The function verifyUnseen in validate.ts adds public keys to the seen set even if they are already in it. This might lead to redundant operations and could be optimized.

chrypnotoad:
```
Summary of all failing tests
FAIL test/unit/src/validateJoinRequest.test.ts (21.152 s)
  ● verifyJoinRequestSigner › should fail with wrong signature owner

    expect(received).toStrictEqual(expected) // deep equality

    Expected: {"fatal": true, "reason": "Bad signature, sign owner and node attempted joining mismatched", "success": false}
    Received: null

      124 |         expect(result).toBeNull()
      125 |       } else {
    > 126 |         expect(result).toStrictEqual(test.expectedError)
          |                        ^
      127 |       }
      128 |     })
      129 |   }

      at Object.<anonymous> (test/unit/src/validateJoinRequest.test.ts:126:24)

  ● verifyJoinRequestSigner › should fail with wrong node public key

    expect(received).toStrictEqual(expected) // deep equality

    Expected: {"fatal": true, "reason": "Bad signature, sign owner and node attempted joining mismatched", "success": false}
    Received: null

      124 |         expect(result).toBeNull()
      125 |       } else {
    > 126 |         expect(result).toStrictEqual(test.expectedError)
          |                        ^
      127 |       }
      128 |     })
      129 |   }

      at Object.<anonymous> (test/unit/src/validateJoinRequest.test.ts:126:24)
 ```
 two of the new tests failed in the PR check
 https://github.com/shardeum/shardus-core/actions/runs/10410903643/job/28833708776?pr=244

**Review by magonjr — COMMENTED** [review]
PR LGTM 
two tests are failing for me 
https://github.com/shardeum/shardus-core/pull/244/files#diff-6691c51708538fc5b3d636534da4165888f4c5746c7fce110a0c467bc02733dfR329

**Review by chrypnotoad — APPROVED** [review]

**Comment by BelfordZ** [comment]
Can you point out the circular dependency that existed, and where it was removed?

I noticed you didnt have any commits that add tests without refactoring. Which test was not possible without needing a refactor?

**Comment by muni-corn** [comment]
> Can you point out the circular dependency that existed, and where it was removed?
> 
> I noticed you didnt have any commits that add tests without refactoring. Which test was not possible without needing a refactor?

These tests required mocking fields and functions that were all included in the root of the `Join` module. Because the actual functions we wanted to test were *also* in the root module, this sort of self-dependency would upset mocking with `jest`. (Because the functions' actual implementations are imported as-is before `jest` can replace any calls with mock implementations.)

For example, all tests using logging functions like `info`, `warn`, or `error` fail with their actual implementations because `p2pLogger` is uninitialized (and I haven't found a decent way of actually initializing it with `log4js` yet; it felt unnecessary).

It also became impossible to change `allowBogon` or `seen` as needed.

So...

- The `logging` module was created to separate the logging functions so that `jest` can mock them. Their mocked implementations currently do nothing when they are called. We can add a custom implementation to use e.g. `console.log`.
- The `state` module was created along with functions `getAllowBogon` and `getSeen` so that we can mock the values of these state variables with `jest` functions. Some code using these variables has been changed to use the functions instead, so that any mock implementations are used.
- The `types` module was created to house TypeScript types specific to the Join module so that other (sub)modules don't have to import the entirety of `index.ts` to use an interface.
- The `validate` module was created to isolate the functions we want to test. *Probably* not necessary with `logging` and `state` separated, but considerably reduces the size of our `Join/index.ts` file, which is nice :)

This fixes any tests that depended on `allowBogon`, `seen`, `info()`, `warn()`, or `error()` (e.g. the IP address tests). The issue was the self-dependency of the root module. The goal of the refactor was to separate any mocked functions (and their dependencies, if necessary) into their own modules so that the modules can be mocked separately from the root module.

File: test/unit/src/validateJoinRequest.test.ts [hunk 2]
```diff
@@ -0,0 +1,441 @@
...
+          ...validJoinRequest.sign,
+          owner: 'already seen',
+        },
+      },
+      expectedError: ALREADY_SEEN_ERROR,
+    },
+    {
+      it: 'should fail if joining node is known by public key',
+      before: () => {
+        ;(getByPubKeyMap as jest.Mock).mockReturnValueOnce(new Map([['known node', {}]]))
```
BelfordZ, line 230: god help us all if we need this semi colon stuff.

PR #246, opened 2024-08-19
external endpoint timeout guard

S0naliThakur:
Linear: https://linear.app/shm/issue/PRI-41
Summary: Added middleware to enforce a timeout on external endpoints, preventing waits beyond a set time limit.

### DIFF
--- a/src/network/index.ts
+++ b/src/network/index.ts
@@ -171,6 +171,8 @@ export class NetworkClass extends EventEmitter {
         let timer = setTimeout(() => {
           if (!res.headersSent) {
             res.status(408).json({ error: 'Request Timeout' })
+            req.timedOut = true
+            req.destroy() // Abort the request processing
           }
         }, timeout)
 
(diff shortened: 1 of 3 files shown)


src/network/index.ts
```diff
@@ -162,11 +166,33 @@ export class NetworkClass extends EventEmitter {
         next()
       }
 
+      const timeoutMiddleware = function (req, res, next) {
+        const timeout = config.network.requestTimeout * 1000 || 5000
```
jintukumardas, line 170: We can remove the  `|| 5000` option.

jintukumardas (changes requested):

src/network/index.ts
S0naliThakur: network in config is optional. 5000 is fallback

jintukumardas (dismissed):

kgmyatthu:
This can be done in much simpler way. Please use express's builtin support for this. Something like this is all you need.
Example: 

```javascript
const server = app.listen()
server.setTimeout(5000)
```
Current implementation in PR add unnecessary LOC(s). Please modify.

kgmyatthu (commented):
Please check the closing statement

PR #253, opened 2024-08-21
RED-103: use nodelistFromStates function in all cycle tx sendGossip calls

mhanson-github:
New PR for this bug.

### DIFF
--- a/src/p2p/CycleAutoScale.ts
+++ b/src/p2p/CycleAutoScale.ts
@@ -15,6 +15,7 @@ import { enterRecovery, enterSafety } from './Modes'
 import { getOurNodeIndex } from './Utils'
 import { shardusGetTime } from '../network'
 import { Utils } from '@shardus/types'
+import { nodeListFromStates } from './Join'
 
 /** STATE */
 
@@ -54,7 +55,7 @@ const gossipScaleRoute: P2P.P2PTypes.GossipHandler<P2P.CycleAutoScaleTypes.Signe
 
     const added = addExtScalingRequest(payload)
     if (!added) return
-    Comms.sendGossip('scaling', payload, tracker, sender, NodeList.byIdOrder, false, 2)
+    Comms.sendGossip('scaling', payload, tracker, sender, nodeListFromStates(['active', 'ready', 'syncing']), false, 2)
   } finally {
     profilerInstance.scopedProfileSectionEnd('gossip-scaling')
   }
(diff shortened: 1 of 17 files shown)


dnlbui:
Ready for review. Rebased and added a two commits to address mistakes ([8ffe854](https://github.com/shardeum/shardus-core/pull/253/commits/8ffe854c493d6577f46d93be739beb1735247418) and [8dde130](https://github.com/shardeum/shardus-core/pull/253/commits/8dde130e9211e650760c8d4cb44be483ccdfc4bb)

src/p2p/CycleAutoScale.ts
```diff
@@ -54,7 +55,19 @@ const gossipScaleRoute: P2P.P2PTypes.GossipHandler<P2P.CycleAutoScaleTypes.Signe
...
+      tracker,
+      sender,
+      nodelistFromStates([
+        P2P.P2PTypes.NodeStatus.ACTIVE,
+        P2P.P2PTypes.NodeStatus.READY,
+        P2P.P2PTypes.NodeStatus.SYNCING,
+      ]),
+      false,
+      2
+    )
```
ahmxdiqbal: only pass active to nodelist from states here

src/p2p/CycleAutoScale.ts
```diff
@@ -131,7 +144,19 @@ function _requestNetworkScaling(upOrDown) {
...
+      '',
+      null,
+      nodelistFromStates([
+        P2P.P2PTypes.NodeStatus.ACTIVE,
+        P2P.P2PTypes.NodeStatus.READY,
+        P2P.P2PTypes.NodeStatus.SYNCING,
+      ]),
+      true,
+      2
+    )
```
ahmxdiqbal: only pass active to nodelist from states here

src/p2p/Join/routes.ts
```diff
@@ -223,14 +223,18 @@ const unjoinRoute: P2P.P2PTypes.Route<Handler> = {
...
+      unjoinRequest,
+      '',
+      null,
+      nodelistFromStates([
+        P2P.P2PTypes.NodeStatus.ACTIVE,
+        P2P.P2PTypes.NodeStatus.READY,
+        P2P.P2PTypes.NodeStatus.SYNCING,
+      ]),
+      true
+    )
```
ahmxdiqbal: whats up with this change? why was it made? it'll probably break unjoin

dnlbui: You're right. I was looking at `serviceQueue-impl` branch to double check this earlier. It has gossip still. I just checked dev (which I should have just referenced this diff). Changed back and resolving this comment after committing.

PR #258, opened 2024-08-25
GOLD-211: ECDSA Multi-sig feature

arhamj:
Task: https://linear.app/shm/issue/GOLD-211/

- Added config to hold a mapping of allowed keys for multi-sig
- Adding and unifying multi-sig related flags.
- Adding support to accept a `verifyMultiSigs` method implementation from the app layer. To let curve specific logic to reside on the app layer.
- Updated and renamed middleware to use the above.
- Updated /df endpoint to use this middleware by default


src/debug/config.ts
```diff
@@ -23,46 +23,45 @@ export function isServiceMode(): boolean {
...
-  if (pkClearance !== undefined) return key
-  return null
+  return getDevPublicKeys()[key] !== undefined ? key : null
 }
 
 export function getDevPublicKeyMaxLevel(clearance?: DevSecurityLevel): string | null {
   const devPublicKeys = getDevPublicKeys()
   let maxLevel = -Infinity
   let maxKey = null
-  for (const key in devPublicKeys) {
```
arhamj: Revert this

arhamj: And rest in this file

jintukumardas: reverted

afostr (approved):
Looks great.   local network smoke test passed.

PR #259, opened 2024-08-26
BLUE-187: Implemenet new POQo receipt types

Glitch18:
https://linear.app/shm/issue/BLUE-187/poqo-receipt-and-ancillary-service-updates

Refactors POQo new reciept across the codebase

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -213,6 +213,9 @@ class Crypto {
   }
 
   hash(obj: HashableObject): string {
+    if (typeof obj === 'string') {
+      return crypto.hash(obj)
+    }
     if (!obj.sign) {
       return crypto.hashObj(obj)
     }
--- a/src/state-manager/TransactionConsensus.ts
+++ b/src/state-manager/TransactionConsensus.ts
@@ -1988,16 +1988,16 @@ class TransactionConsenus {
    *
    * @param queueEntry
    */
-  hasAppliedReceiptMatchingPreApply(queueEntry: QueueEntry, appliedReceipt: AppliedReceipt): boolean {
+  hasAppliedReceiptMatchingPreApply(queueEntry: QueueEntry, signedReceipt: SignedReceipt): boolean {
     if (queueEntry.preApplyTXResult == null || queueEntry.preApplyTXResult.applyResponse == null) {
       /* prettier-ignore */ if (logFlags.debug) this.mainLogger.debug(`hasAppliedReceiptMatchingPreApply  ${queueEntry.logID} preApplyTXResult == null or applyResponse == null`)
       return false
     }
     // This is much easier than the old way
     if (queueEntry.ourVote) {
-      const receipt = queueEntry.appliedReceipt2 ?? queueEntry.recievedAppliedReceipt2
+      const receipt = queueEntry.signedReceipt
       if (receipt != null && queueEntry.ourVoteHash != null) {
-        const receiptVoteHash = this.calculateVoteHash(receipt.appliedVote)
+        const receiptVoteHash = this.calculateVoteHash(receipt.proposal)
         if (receiptVoteHash === queueEntry.ourVoteHash) {
           return true
         } else {
@@ -2008,7 +2008,7 @@ class TransactionConsenus {
       return false
     }
 
-    if (appliedReceipt == null) {
+    if (signedReceipt == null) {
       return false
     }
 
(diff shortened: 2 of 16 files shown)


src/crypto/index.ts
```diff
@@ -213,6 +213,9 @@ class Crypto {
   }
 
   hash(obj: HashableObject): string {
+    if (typeof obj === 'string') {
```
afostr, line 216: I am curious about the reason for this change.

src/state-manager/TransactionConsensus.ts
```diff
@@ -1683,16 +1679,16 @@ class TransactionConsenus {
...
-              /* prettier-ignore */ if (logFlags.error) this.mainLogger.error(`binary_poqo_send_vote mismatch txGroupCycle for txid: ${readableReq.txid}, sender's txGroupCycle: ${readableReq.txGroupCycle}, our txGroupCycle: ${queueEntry.txGroupCycle}`)
-              nestedCountersInstance.countEvent(
-                'poqo',
-                'binary_poqo_send_vote: mismatch txGroupCycle for tx ' + readableReq.txid
-              )
-            }
-            delete readableReq.txGroupCycle
-          }
+          // if (readableReq.txGroupCycle) {
+          //   if (queueEntry.txGroupCycle !== readableReq.txGroupCycle) {
```
afostr, line 1683: did we loose txGroupCycle?   I think that was developed after the initial Poqo receipt spec

src/state-manager/TransactionConsensus.ts
```diff
@@ -1714,40 +1710,53 @@ class TransactionConsenus {
...
+      voteTime: 0
     }
 
-    let validSignatures = 0;    
-    for (const owner of ownerToSignMap.keys()) {
-      const signedObject = { ...appliedVoteHash, sign: ownerToSignMap.get(owner) };
-      if (this.crypto.verify(signedObject, owner)) {
+    for (let i = 0; i < receipt.signaturePack.length; i++) {
+      const sign = receipt.signaturePack[i]
+      if (!executionGroupNodes.has(sign.owner)) continue
```
afostr, line 1743: we need a set to ensure that signers are unique

afostr: I think a dishonest aggregator can just add the same other vote 87 times if we don't check for unique

src/state-manager/TransactionConsensus.ts
```diff
@@ -1714,40 +1710,53 @@ class TransactionConsenus {
...
+      appliedVoteHash.voteTime = receipt.voteOffsets[i]
+      const signedObject = { ...appliedVoteHash, sign };
+      if (this.crypto.verify(signedObject, sign.owner)) {
         validSignatures++;
       }
     }
+
+    const totalNodes = executionGroupNodes.size;
+    const requiredMajority = Math.ceil(totalNodes * this.config.p2p.requiredVotesPercentage)
```
afostr, line 1752: security prefers the base value to use our consensus group size.  It does mean care will need to be taken if someone wants to run a local test network of 20 nodes with say a consensus group size of 128... they will need to adjust or no votes will go through.

src/state-manager/TransactionConsensus.ts
```diff
@@ -3575,183 +3574,207 @@ class TransactionConsenus {
...
       //append our vote
       appliedVoteHash = this.crypto.sign(appliedVoteHash)
-      if (this.stateManager.transactionQueue.useNewPOQ === false)
-        this.tryAppendVoteHash(queueEntry, appliedVoteHash)
+      // if (this.stateManager.transactionQueue.useNewPOQ === false)
+      this.tryAppendVoteHash(queueEntry, appliedVoteHash)
 
       // save our vote to our queueEntry
+      this.crypto.sign(ourVote)
+      this.crypto.sign(proposal)
```
afostr, line 3617: do we need to sign the proposal?

src/state-manager/TransactionConsensus.ts
```diff
@@ -3575,183 +3574,207 @@ class TransactionConsenus {
...
+        applied: appliedVote.transaction_result,
+        cantApply: appliedVote.cant_apply
       }
       const stateHash = {
-        account_id: vote.account_id,
-        account_state_hash_after: vote.account_state_hash_after,
-        account_state_hash_before: vote.account_state_hash_before,
+        account_id: appliedVote.account_id,
+        account_state_hash_after: appliedVote.account_state_hash_after,
+        account_state_hash_before: appliedVote.account_state_hash_before,
```
afostr, line 3774: below where we setup votToHash the structure looks different than for an applied TX.  please double check and confirm if this is ok

src/state-manager/TransactionConsensus.ts
```diff
@@ -3760,18 +3783,19 @@ class TransactionConsenus {
...
-        txId: vote.txid,
-        transaction_result: vote.transaction_result,
-        account_id: vote.account_id,
-        account_state_hash_after: vote.account_state_hash_after,
-        account_state_hash_before: vote.account_state_hash_before,
-        cant_apply: vote.cant_apply,
-        app_data_hash: vote.app_data_hash,
+        txId: appliedVote.txid,
+        transaction_result: appliedVote.transaction_result,
+        account_id: appliedVote.account_id,
```
afostr, line 3790: useNewPOQ case can be commented out along with the non POQo branch

src/state-manager/TransactionConsensus.ts
```diff
@@ -4009,133 +4034,133 @@ class TransactionConsenus {
...
+    /* prettier-ignore */ if (logFlags.debug) this.mainLogger.debug(`tryAppendVote collectedVotes: ${queueEntry.logID}   vote: ${utils.stringifyReduce(vote)}`)
 
-        if (currentVote.sign.owner === vote.sign.owner) {
-          // already in our list so do nothing and return
-          return false
-        }
-      }
-      queueEntry.lastVoteReceivedTimestamp = shardusGetTime()
+    // just add the vote if we dont have any yet
+    if (numVotes === 0) {
```
afostr, line 4044: does the first vote bypass security if some other node sends it to us before we make our first vote?

afostr: I think we should be checking that the vote is valid/legal before we append it and return from this func.  otherwise it allows gossip to multiply DOS

afostr: please correct me if I missed a layer of security before this

src/state-manager/TransactionQueue.ts
```diff
@@ -5348,13 +5347,24 @@ class TransactionQueue {
...
     queueEntry.appliedReceiptForRepair2 = null
 
     delete queueEntry.recievedAppliedReceipt2
     delete queueEntry.appliedReceiptForRepair2
 
+    queueEntry.signedReceipt = 
+      queueEntry.signedReceipt ??
+      queueEntry.receivedSignedReceipt ??
+      queueEntry.signedReceiptForRepair ??
+      queueEntry.signedReceiptFinal
```
afostr, line 5360: just want to check if there is any priority to signedReceiptFinal, if so maybe should be at the front of the chain?

src/state-manager/TransactionQueue.ts
```diff
@@ -6541,98 +6547,92 @@ class TransactionQueue {
...
+              // }
 
               /* prettier-ignore */ if (logFlags.debug) this.mainLogger.debug(`processAcceptedTxQueue2 tryProduceReceipt result : ${queueEntry.logID} ${utils.stringifyReduce(result)}`)
 
               //todo this is false.. and prevents some important stuff.
               //need to look at appliedReceipt2
-              if (this.stateManager.getReceipt2(queueEntry) != null) {
-                const receipt2 = this.stateManager.getReceipt2(queueEntry)
+              if (this.stateManager.getSignedReceipt(queueEntry) != null) {
+                const signedReceipt = this.stateManager.getSignedReceipt(queueEntry)
```
afostr, line 6635: the old code did the same thing twice, but I think we don't need to here

afostr (commented):
my comments were stuck in pending state all night.   :(

src/crypto/index.ts
Glitch18: The lack of this string check was causing an error in the proposal hash function. This hash function allows strings (part of HashableObject) but when actually given a string, it calls crypto-util's hashObj which does not accept strings and throws an error. 

This change allows it to process strings correctly and fix that error

src/state-manager/TransactionConsensus.ts
Glitch18: Oh actually we never reach this else if statement. In the current POQ setup with the new types (Proposal), we will only ever need the first if statement 

`if (this.stateManager.transactionQueue.usePOQo && (vote as Proposal).applied !== undefined)`

But actually, I'll comment out the other else if and else statements since they're pretty much deprecated now.

src/state-manager/TransactionConsensus.ts
Glitch18: We should still have all of the three checks happen for this vote when numVotes === 0. The only thing that we would skip for this vote is checking if we have any existing votes from the same sender. Even if we didn't have this if block the result would be the same as the array that the next for loop iterates over would be empty. So I think having or not having the if statement is pretty much the same.

PR #260, opened 2024-08-27
SEC-338: Limit the proxy handler to only allow a specific route.

mgthuramoemyint:
Limiting handler to only allow a specific route.

### DIFF
--- a/src/p2p/Lost.ts
+++ b/src/p2p/Lost.ts
@@ -271,7 +271,9 @@ export function init() {
             {}
           )
         } else {
-          res = await Comms.ask(targetNode, payload.route, payload.message)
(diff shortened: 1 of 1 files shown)


src/p2p/Lost.ts
```diff
@@ -271,7 +271,9 @@ export function init() {
             {}
           )
         } else {
-          res = await Comms.ask(targetNode, payload.route, payload.message)
+          error(`proxy handler route is not get_trie_hashes`)
```
kun6fup4nd4: can we add what exact route was sent to log message ? without log injection vuln.

mgthuramoemyint: will do

mssabr01 (approved):

PR #262, opened 2024-08-27
Blue-161 Account patcher improvements in radix sharing

thantsintoe:
- do not process the shared radix if its length is different from treeSyncDepth
- this will prevent a bad  node from spreading bad radixes and causing other nodes to be false-positive in OOS detection


jairajdev (dismissed):

afostr (approved):
This looks great

PR #263, opened 2024-08-27
BLUE-230 Remove the checkGossipPayload from the joinarchiver gossip handler, addArchiverJoinRequest would take care of validating archiver join request payload

jairajdev:


thantsintoe:
Looks good. `addArchiverJoinRequest` is validating the payload already.

PR #271, opened 2024-09-03
BLUE-252 use Set<> to only allow unique voters to add a vote for radix

thantsintoe:


kun6fup4nd4 (dismissed):
minor - why do we need counter if can use .size everywhere ?

afostr (approved):
looks good

PR #272, opened 2024-09-03
comment out deprecated and unused "gossip-final-state" handler

thantsintoe:
- new POQo gossip endpoints and CT handlers properly verify the sender of the receipt or the final data.


afostr (approved):
looks good

PR #275, opened 2024-09-03
Fix compare_cert AJV check

arhamj:
https://linear.app/shm/issue/GOLD-254/

### DIFF
--- a/src/types/ajv/CompareCert.ts
+++ b/src/types/ajv/CompareCert.ts
@@ -104,4 +104,5 @@ function addSchemaDependencies(): void {
 // Function to register the schema
 function addSchemas(): void {
   addSchema(AJVSchemaEnum.CompareCertReq, schemaCompareCert)
+  addSchema(AJVSchemaEnum.CompareCertResp, schemaCompareCert)
 }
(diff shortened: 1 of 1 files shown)


src/types/ajv/CompareCert.ts
```diff
@@ -80,21 +80,15 @@ const schemaCompareCert = {
...
         'nodeListHash',
         'archiverListHash',
         'standbyNodeListHash',
-        'random',
       ],
       additionalProperties: true,
     },
   },
-  required: ['certs', 'record'],
+  required: ['certs'],
```
jintukumardas, line 90: is this accidental or we don't need the record to be the required param?

src/types/ajv/CompareCert.ts
```diff
@@ -110,4 +104,5 @@ function addSchemaDependencies(): void {
 // Function to register the schema
 function addSchemas(): void {
   addSchema(AJVSchemaEnum.CompareCertReq, schemaCompareCert)
+  addSchema(AJVSchemaEnum.CompareCertResp, schemaCompareCert)
```
jintukumardas, line 107: Added missing schema for CompareCertResp

PR #278, opened 2024-09-03
PRI 122: Replace res.send with res.json

urnotsam:


src/logger/index.ts
```diff
@@ -524,15 +524,14 @@ class Logger {
         } else {
           const sanitizedFileName = path.basename(requestedFileName)
           if (!validFileNames.includes(sanitizedFileName)) {
-            res.status(400).send('Invalid log file specified')
+            res.status(400).json('Invalid log file specified')
```
ahmxdiqbal: this is res.json, but we are sending a string instead of a json object. sam is out, so i'll push the fix soon

src/p2p/Archivers.ts
```diff
@@ -900,19 +900,19 @@ export function registerRoutes() {
     const accepted = await addArchiverJoinRequest(joinRequest)
     if (!accepted.success) {
       warn('Archiver join request not accepted.')
-      return res.send(
+      return res.json(
         Utils.safeStringify({ success: false, error: `Archiver join request rejected! ${accepted.reason}` })
```
ahmxdiqbal: similar thing here, no need to stringify for res.json

src/p2p/Archivers.ts
```diff
@@ -921,12 +921,12 @@ export function registerRoutes() {
     const accepted = await addLeaveRequest(leaveRequest)
     if (!accepted.success) {
       warn('Archiver leave request not accepted.')
-      return res.send(
+      return res.json(
         Utils.safeStringify({ success: false, error: `Archiver leave request rejected! ${accepted.reason}` })
```
ahmxdiqbal: same thing

src/p2p/LostArchivers/routes.ts
```diff
@@ -309,7 +309,7 @@ const reportFakeLostArchiverRoute: P2P.P2PTypes.Route<Handler> = {
     const archiver = publicKey ? getArchiverWithPublicKey(publicKey) : getRandomAvailableArchiver()
     if (archiver) {
       funcs.reportLostArchiver(archiver.publicKey, 'fake lost archiver report')
-      res.send(
+      res.json(
         Utils.safeStringify(
```
ahmxdiqbal: same thing

src/p2p/LostArchivers/routes.ts
```diff
@@ -320,7 +320,7 @@ const reportFakeLostArchiverRoute: P2P.P2PTypes.Route<Handler> = {
         )
       )
     } else {
-      res.send(
+      res.json(
         Utils.safeStringify(
```
ahmxdiqbal: same thing

src/logger/index.ts
```diff
@@ -486,36 +486,36 @@ class Logger {
           if (err) {
             // Handle error (e.g., file not found)
             console.error('Error reading file:', err)
-            res.status(500).send('Error reading file')
+            res.status(500).json({ error: 'Error reading file' })
```
afostr, line 489: I think in the cases where we converted strings to json we could have posted:
res.type('txt')
in front of the call instead.
I am a bit concerned that there could be parts of the pipeline or tools that could break if we change the schema

afostr (approved):
looks good

PR #279, opened 2024-09-04
red-268: add cycle to  unjoin request

magonjr:


src/p2p/Join/v2/unjoin.ts
```diff
@@ -72,6 +73,14 @@ export function validateUnjoinRequest(unjoinRequest: SignedUnjoinRequest): Resul
     return err(new Error(`unjoin request from ${unjoinRequest.publicKey} already exists`))
   }
 
+   // cycle number check
+   const cycleNumber = CycleChain.getNewest().counter
+   if (cycleNumber !== unjoinRequest.cycleNumber) {
+     return err(
```
mhanson-github: One suggestion I have here is to possibly soften this check while keeping the overall protection.

Maybe check for +/- within 1-2 cycles of current cycle.

Concern of shooting for an exact match may cause some unjoins to not be removed.

mhanson-github (changes requested):

afostr (approved):
looks good

PR #285, opened 2024-09-05
RED-261: add cycleRecord validation

PudgyPug:


mssabr01 (dismissed):

src/shardus/shardus-types.ts
```diff
@@ -973,6 +973,8 @@ export interface ServerConfiguration {
     requiredVotesPercentage: number
     // /** a fix to prevent node from producing different ts for same txId */
     timestampCacheFix: boolean
+    // whether to use ajv for validation for CycleRecord sync
+    useAjvCycleRecordValidation: boolean
```
afostr, line 978: I think this may not be ready to merge yet.   I don't see this flag initialized or checked.

mssabr01:
I see the new schema but I don't see a .compile() or validate used anywhere. I don't believe we are actually using the schema for a fix here

mssabr01 (changes requested):
I see the new schema but I don't see a .compile() or validate used anywhere. I don't believe we are actually using the schema for a fix here

src/shardus/shardus-types.ts
PudgyPug: Sorry I missed the `Sync.ts` file commit

mssabr01 (dismissed):

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,284 @@
...
+      items: {
+        type: 'object',
+        properties: {
+          cycle: { type: 'number' },
+          hash: { type: 'string' },
+        },
+      },
+    },
+  },
+  required: [
```
afostr, line 270: it seems like there should be more items in the required list here?  are the missing fields actually safe to ignore.
Also AJV should have a setting to reject if there are extra fields getting injected

mhanson-github (dismissed):
Andrew live-reviewed this on call with me.

src/types/ajv/CycleRecordSchema.ts
PudgyPug: I can add more fields that should in theory never be null to the required fields however my concern is that there’s edge cases that do have null fields and since sync of cycle records is pretty central I think that’d be a bit risky. current validation isn’t checking for other fields than the ones I included in the schema either. 
Should we wait for the RED-300 task before adding more required fields (just to be sure)?

PudgyPug: added all the fields that should in theory be there now. if this fails us we can turn off the AJV validation flag still

devendra-shardeum:
> I see the new schema but I don't see a .compile() or validate used anywhere. I don't believe we are actually using the schema for a fix here

@mssabr01 it's being called inside like 
1. step 1:  verifyPayload.. : https://github.com/shardeum/shardus-core/blob/dev/src/types/ajv/Helpers.ts#L105C1-L113C1
2. step 2 : getVerifyFunction : https://github.com/shardeum/shardus-core/blob/dev/src/utils/serialization/SchemaHelpers.ts#L32C1-L42C18

src/p2p/Sync.ts
```diff
@@ -551,154 +552,171 @@ export function totalNodeCount(cycle: P2P.CycleCreatorTypes.CycleRecord) {
...
-        warn('Validation failed for cycleRecord.joined: ' + err)
-        return false
-      }
-    }
-    for (const nodeId of cycleRecord.returned) {
-      if (typeof nodeId !== 'string') {
-        warn('Validation failed for cycleRecord.returned: ' + err)
-        return false
-      }
+  if (config.p2p.useAjvCycleRecordValidation) {
```
devendra-shardeum: why are we using config as a toggle for ajv validation here ?

src/p2p/Sync.ts
```diff
@@ -551,154 +552,171 @@ export function totalNodeCount(cycle: P2P.CycleCreatorTypes.CycleRecord) {
...
+    const errors = verifyPayload(AJVSchemaEnum.CycleRecords, cycles)
```
devendra-shardeum: whenever we validate an array of object with ajv, it should be iterated through the items it contains and then whatever is in correct or expected format should be marked as valid and returned ( or should be added to the final result array which needs to be returned at the end .. in this case validationCycles..)..also in parent function ( get-cycles ) this is marked as // [TODO] Validate whatever came in  .. and the part of force casting can run into runtime-exceptions.. so validation needs to be done before that.. .and verifyPayload must be surrounded by try/catch.. and also please add the try/catch to the parent function as well.. i.e. get-cycles

src/p2p/Sync.ts
PudgyPug: I've talked with Andrew about this. The reason for the toggle is cycleRecord validation is a central part and in case we are too strict with our schema (e.g. some weird edge case) that would cause cycle records to not become valid anymore we will want to switch back. 
This should be phased out at a later point when we are more confident with this (can be done with RED-300 - should add a test by taking the cycle records from a long running network ideally ITN3)

src/p2p/Sync.ts
PudgyPug: I would argue that returning partial valid data here could cause to more issues later down the line. So if we validate a bunch of cycle records 1 by 1 maybe one of them is failing but that could lead to invalid node lists and now with network generated txs that could lead to missed txs.

PudgyPug: I'll add a try/catch to this. What do you think of doing the try/catch in `verifyPayload` function itself? we already return an Error if something through validation. We could handle that the same way

devendra-shardeum: consider a scenario.. where i have a single malicious node.. that always returns.. bad or incorrect format in getcyles ( query .. i.e. sync-cycles ).. so this will result in ajv validation being failed every time.. just because of 1 malicious active node present in the network.. and the sync-cycles step would never succeed. @PudgyPug

PudgyPug: > and the part of force casting can run into runtime-exceptions

could you elaborate on this? Casting in typescript is not causing runtime exceptions

PudgyPug: > consider a scenario.. where i have a single malicious node.. that always returns.. bad or incorrect format in getcyles ( query .. i.e. sync-cycles ).. so this will result in ajv validation being failed every time.. just because of 1 malicious active node present in the network.. and the sync-cycles step would never succeed. @PudgyPug

robustQuery is taking care of that

PudgyPug: cleared up over huddle. try/catch has been added to validation

devendra-shardeum: based on the conversations in dm.. and also as per my observations ( we are using robsutQuery to query the nodes )  : Redundancy is the minimum number of consistent responses needed, not the majority of the entire network. robustQuery doesn’t query all nodes in the network at once. Instead, it queries a subset of nodes and looks for redundancy within that subset and case where [ validCycleRecord, notValidCycleRecord ] will be present will be less likely because of the no. of good nodes present in the network will be more.. so adding try/catch should suffice in this case beside the ajv validation with response ( in the parent function getCycles )..

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,346 @@
...
+        properties: {
+          publicKey: { type: 'string' },
+          ip: { type: 'string' },
+          port: { type: 'number' },
+          curvePk: { type: 'string' },
+        },
+        required: ['publicKey', 'ip', 'port', 'curvePk'],
+      },
+    },
+    refreshedConsensors: {
```
devendra-shardeum, line 34: this should be corrected i think.. refreshedConsensors would correct when the items in object i.e. node have the following as it's properties.. for eg. we don't see curve public key and status in it's properties. : 
image

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,346 @@
...
+    'marker',
+    'maxSyncTime',
+    'nodeListHash',
+    'refuted',
+    'refutedArchivers',
+    'removed',
+    'removedArchivers',
+    'returned',
+    'standby',
+    'standbyAdd',
```
devendra-shardeum: standbyAdd is optional

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,346 @@
...
+    'standbyNodeListHash',
+    'standbyRefresh',
+    'standbyRemove',
```
devendra-shardeum: standbyRefresh and standbyRemove are optional

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,346 @@
...
+    'startedSyncing',
```
devendra-shardeum: startedSyncing is optional

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,346 @@
...
+    'refreshedArchivers',
+    'refreshedConsensors',
+    'activated',
+    'activatedPublicKeys',
+    'active',
+    'apoptosized',
+    'appRemoved',
+    'archiverListHash',
+    'expired',
+    'finishedSyncing',
```
devendra-shardeum: finishedSyncing is optional

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,346 @@
...
+    'joined',
+    'joinedArchivers',
+    'joinedConsensors',
+    'leavingArchivers',
+    'lost',
+    'lostAfterSelection',
```
devendra-shardeum: lostAfterSelection is optional

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,345 @@
...
+              },
+              version: { type: 'string' },
+            },
+            required: ['mustUseAdminCert', 'stakeCert', 'version'],
+          },
+          cycleMarker: { type: 'string' },
+          nodeInfo: {
+            type: 'object',
+            properties: {
+              publicKey: { type: 'string' },
```
devendra-shardeum, line 154: node info is missing fields .. activeTimestamp, syncingTimestamp, refreshed Counter is also part of node info but not required.. please use the below for reference : 
image

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,366 @@
...
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          cycle: { type: 'number' },
+          hash: { type: 'string' },
+        },
+      },
+    },
+    networkReceiptHash: {
```
devendra-shardeum, line 221: networkReceiptHash needs to be added in required.

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,366 @@
...
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          cycle: { type: 'number' },
+          hash: { type: 'string' },
+        },
+      },
+    },
+    networkSummaryHash: {
```
devendra-shardeum, line 231: networkSummaryHash needs to be required field

src/types/ajv/CycleRecordSchema.ts
```diff
@@ -0,0 +1,366 @@
...
+    },
+    standbyNodeListHash: { type: 'string' },
+    standbyRefresh: { type: 'array', items: { type: 'string' } },
+    standbyRemove: { type: 'array', items: { type: 'string' } },
+    startedSyncing: { type: 'array', items: { type: 'string' } },
+    syncing: { type: 'number' },
+    target: { type: 'number' },
+    desired: { type: 'number' },
+    random: { type: 'number' },
+    networkDataHash: {
```
devendra-shardeum, line 211: networkDataHash needs to be a required field

src/types/ajv/CycleRecordSchema.ts
PudgyPug: not required in practice

src/types/ajv/CycleRecordSchema.ts
PudgyPug: not required in practice

src/types/ajv/CycleRecordSchema.ts
PudgyPug: not required in practice

devendra-shardeum (dismissed):

urnotsam (approved):

ahmxdiqbal (approved):

PR #288, opened 2024-09-10
RED-265/BB-33152: shardus-core - switch node to 18.19.1

ahmxdiqbal:


ahmxdiqbal:
I would've liked to have also updated package-lock.json, but since shardus core doesn't let you compile if the specified node engines mismatch between shardus-core and its dependencies, I think we need to wait until new dist are released for the dependencies before updating types

urnotsam (approved):

PR #294, opened 2024-09-12
GOLD-265 -- feat: archiver socketio now require a timestamped sig for replay prevent

kgmyatthu:


src/shardus/index.ts
```diff
@@ -93,6 +93,7 @@ import SocketIO from 'socket.io'
 import { nodeListFromStates, queueFinishedSyncingRequest } from '../p2p/Join'
 import * as NodeList from '../p2p/NodeList'
 import { P2P } from '@shardus/types'
+import { Socket } from 'net'
```
arhamj: unused?

arhamj (dismissed):
LGTM. Minor comment.

src/shardus/index.ts
kgmyatthu: You're right, I mis-imported the wrong socket from net. (meant to use the one from socketio) and forget to delete it after.

arhamj (dismissed):

S0naliThakur (approved):

PR #296, opened 2024-09-13
fix: check ip of archiver against archiver map node maintains

ahmxdiqbal:


src/shardus/index.ts
```diff
@@ -492,26 +492,31 @@ class Shardus extends EventEmitter {
           }
 
           const archiverIP = socket.handshake.address.split('::ffff:').pop();
-          if (!utils.isValidIPv4(archiverIP)) {
-            this.mainLogger.error(`❌ Invalid IP-Address of Archiver: ${archiverIP}`)
-            socket.disconnect()
-            return
-          }
+          console.log('FOR DEBUG PURPOSES: our IP', Self.ip)
+          console.log('FOR DEBUG PURPOSES: archiverIP', archiverIP)
```
mssabr01: These don't need to make it into dev

src/shardus/index.ts
```diff
@@ -492,26 +492,31 @@ class Shardus extends EventEmitter {
...
+          //   return
+          // }
           const archiverCreds = JSON.parse(socket.handshake.query.data)
           const isValidSig = this.crypto.verify(archiverCreds, archiverCreds.publicKey)
           if (!isValidSig) {
-            this.mainLogger.error(`❌ Invalid Signature from Archiver @ ${archiverIP}`)
+            this.mainLogger.error(`❌ Invalid Signature from Archiver`)
             socket.disconnect()
             return
           }
```
mssabr01, line 514: Crypto.verify is expensive so move this to be the last check. Also use the public key from the archiver map, not the public key included in the socket data. Either use the public key from `Archivers` map directly, or make sure the public key included in the request matches the one for that specific archiver in the `Archivers` map

mssabr01 (changes requested):

src/shardus/index.ts
ahmxdiqbal: yeah for sure

src/shardus/index.ts
chrypnotoad: just to confirm is this what you are asking for @mssabr01 ?
1. get public key from data
2. get matching archiver from Archivers.recipients with key from data (fail if no matching archiver)
3. get key from recipient
4. validate signature using recipient pubkey instead of pubkey from data

chrypnotoad: it is mapped by public key
```
const recipient = Archivers.recipients.get(archiverCreds.publicKey)
```
https://github.com/shardeum/shardus-core/blob/dev/src/p2p/Archivers.ts#L47-L50

PR #299, opened 2024-09-16
SHARD-460 : Add new p2p config 'payloadSizeLimitInBytes'

jintukumardas:


src/network/index.ts
```diff
@@ -197,6 +188,8 @@ export class NetworkClass extends EventEmitter {
         hashKey: this.shardusCryptoHashKey,
         signingSecretKeyHex: this.signingSecretKeyHex,
       },
+      payloadSizeLimitInBytes: config.p2p.payloadSizeLimitInBytes,
```
abdulazeem-tk4vr: might be nice to have this under a config object given that other variables with similarity in use case are also under an object structure for the same function definition .

jintukumardas: Yes, it makes sense. It is updated, thanks!

aniketdivekar (dismissed):

S0naliThakur (dismissed):

abdulazeem-tk4vr (dismissed):

aniketdivekar (approved):

abdulazeem-tk4vr (approved):

PR #305, opened 2024-09-24
BLUE-179 Limiting archivers in the network feature using allowed archivers list

jairajdev:
- Restrict archivers in the network by adding allowed archivers list in the network config
- Limiting archivers is put behind the allowOnlyApprovedArchivers config
- Added pruneDisallowedArchivers feature to remove active archivers that are not in the new allowed archivers list


thantsintoe (dismissed):
The changes are good.

src/p2p/Archivers.ts
```diff
@@ -311,6 +320,10 @@ export function addArchiverJoinRequest(joinRequest: P2P.ArchiversTypes.Request,
       return { success: false, reason: 'This node is not ready to accept this request!' }
     }
   }
+  if (!crypto.verify(joinRequest, joinRequest.nodeInfo.publicKey)) {
+    warn('addJoinRequest: bad signature')
+    return { success: false, reason: 'bad signature ' }
+  }
```
mssabr01, line 331: This is checking the sig against the public key in the request, but we've already checked that the pubkey is in the allowed list and is not already in the active archiver list. This is fine

src/p2p/Archivers.ts
```diff
@@ -425,6 +433,10 @@ export function addLeaveRequest(leaveRequest: P2P.ArchiversTypes.Request, tracke
       reason: 'This archiver leave request timestamp exceeds acceptable timestamp range',
     }
   }
+  if (!crypto.verify(leaveRequest, leaveRequest.nodeInfo.publicKey)) {
```
mssabr01, line 445: Please compare the sig in the leave request to the known public key for the archiver that is requesting to leave. We don't want to verify the sig against an untrusted key that is included in the request. Simply checking that the pub key is present in the active archivers list is not enough, we need to check that the key we know for that archiver is the one that signed the leaveRequest.

src/shardus/shardus-types.ts
```diff
@@ -822,6 +822,14 @@ export interface ServerConfiguration {
...
     /** A node can serve only X max archivers for data transfer and can refuse other archivers requests if it’s serving the max number */
     maxArchiversSubscriptionPerNode: number
+    // /** The desired number of archivers in the network ( which can go up to max_archivers in the network )*/
+    // desiredArchivers: number
+    /** Allow only the allowed archivers to join the network */
+    allowOnlyApprovedArchivers: boolean
+    /** The allowed public keys for archivers that can join the network */
+    allowedArchiversList: string[]
+    /** This is to remove the disallowed archivers that are currently active in the network; which can be used after allowedArchiversList is updated */
+    pruneDisallowedArchivers: boolean
```
mssabr01, line 832: Is there ever a scenario where we would have allowOnlyApprovedArchivers would be true and pruneDisallowedArchivers would be false? I think pruneDisallowedArchivers is redundant and we want the network to always remove unapproved archivers if allowOnlyApprovedArchivers is true

mssabr01 (changes requested):

mssabr01:
Stale

PR #306, opened 2024-09-24
SHARD-888: Remove unused handlers

mgthuramoemyint:


src/state-manager/TransactionQueue.ts
```diff
@@ -718,59 +718,59 @@ class TransactionQueue {
       spreadTxToGroupSyncingBinaryHandler.handler
     )
 
-    this.p2p.registerGossipHandler(
-      'spread_tx_to_group',
```
mssabr01: This needs to remain

mssabr01 (changes requested):

src/state-manager/TransactionQueue.ts
devendra-shardeum: thura corrected the fix and i also couldn't anything else which was still getting used .. with the commented endpoints in the pr.. @mssabr01

devendra-shardeum (dismissed):
changes looks good.. endpoints are not getting used elsewhere.. so safe to comment them..

PR #307, opened 2024-10-01
GOLD-277 : Required minimum for consensus should be STATIC not dynamic number

abdulazeem-tk4vr:
Linear: https://linear.app/shm/issue/GOLD-277/required-minimum-for-consensus-should-be-static-not-dynamic-number

### DIFF
--- a/src/snapshot/index.ts
+++ b/src/snapshot/index.ts
@@ -399,10 +399,9 @@ export async function safetySync() {
   })
 
   // Figure out which nodes hold which partitions in the new network
-  const minNodesPerConsensuGroup = 32
   const nodesPerConsensusGroupUsed = Math.max(
     Context.config.sharding.nodesPerConsensusGroup,
-    minNodesPerConsensuGroup
+    Context.config.sharding.minNodesPerConsensusGroup,
   )
   const shardGlobals = ShardFunctions.calculateShardGlobals(
     safetyNum,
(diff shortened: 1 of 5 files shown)


src/snapshot/index.ts
```diff
@@ -399,10 +399,15 @@ export async function safetySync() {
   })
 
   // Figure out which nodes hold which partitions in the new network
+  const minNodesPerConsensuGroup = 32
```
timadinorth: magic number ? why not have parameter in config with static number ?

abdulazeem-tk4vr: I am a bit confused how this is any different than this in the config. 
```
  sharding: { nodesPerConsensusGroup: 5, nodesPerEdge: 2, executeInOneShard: false },
  ```

abdulazeem-tk4vr: image

abdulazeem-tk4vr: What I meant to say was why cant we set `config.sharding.nodesPerConsensusGroup` to a static value like 32 ?
Why is there a need to create a new variable? 
Do we want 2 config variables? 
- One is `config.sharding.nodesPerConsensusGroup` which is dynamic and can change
- the other is `config.sharding.minNodesPerConsensusGroup` which is constant and set to 32. 

We shall then use the max of these 2 values. Does this sound good?

abdulazeem-tk4vr: I am going to make this change.

PR #308, opened 2024-10-02
GOLD-277 : Required minimum for consensus should be STATIC not dynamic number

abdulazeem-tk4vr:
https://linear.app/shm/issue/GOLD-277/required-minimum-for-consensus-should-be-static-not-dynamic-number

### DIFF
--- a/src/state-manager/TransactionQueue.ts
+++ b/src/state-manager/TransactionQueue.ts
@@ -2207,6 +2207,14 @@ class TransactionQueue {
           //set the nodes that are in the executionGroup.
           //This is needed so that consensus will expect less nodes to be voting
           const unRankedExecutionGroup = homeShardData.homeNodes[0].consensusNodeForOurNodeFull.slice()
+
+          if (txQueueEntry.executionGroup.length < this.config.sharding.minNodesPerConsensusGroup) {
+            nestedCountersInstance.countEvent(
+              'executionGroup',
+              'length is lower than minimumConsensusGroup size'
+            )
+            /* prettier-ignore */ if (logFlags.verbose) if (logFlags.error) this.mainLogger.error(`executionGroup length is lower than minimumConsensusGroup size`)
+          }
           if (this.usePOQo) {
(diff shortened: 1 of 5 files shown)


abdulazeem-tk4vr:
This is the final change that is needed imo. If the execution group length has a lower size, we throw an error else we use the larger value. Because if execution group is not smaller, it will then use the value it has which is larger than the minimumConsensusGroup size.
image
image

abdulazeem-tk4vr:
> This is the final change that is needed imo. If the execution group length has a lower size, we throw an error else we use the larger value. Because if execution group is not smaller, it will then use the value it has which is larger than the minimumConsensusGroup size. image image

The team decided that we only need to make changes where the consensus decisions were being made. There were other places where we share the information after consensus and we concluded that it wouldn't be an issue

src/state-manager/TransactionQueue.ts
```diff
@@ -2182,6 +2182,19 @@ class TransactionQueue {
           //set the nodes that are in the executionGroup.
           //This is needed so that consensus will expect less nodes to be voting
           const unRankedExecutionGroup = homeShardData.homeNodes[0].consensusNodeForOurNodeFull.slice()
+
+          // here we  reject the transaction if the executionGroup.length is lesser than minimumConsensusGroup size, so that we dont have to go ahead with the remaining processing and hereby maintaining the safe node count for consensus
+          if (txQueueEntry.executionGroup.length < this.config.sharding.minNodesPerConsensusGroup) {
```
aniketdivekar, line 2211: @abdulazeem-tk4vr is it possible that execution group length is greater than `minNodesPerConsensusGroup`? If so, is that fine in the premise of this ticket?

abdulazeem-tk4vr: yes that is fine.

aniketdivekar:
closing in favour of this [PR](https://github.com/shardeum/shardus-core/pull/339)

PR #310, opened 2024-10-14
fix: add duplicate signature check in verifyAppliedReceipt to prevent…

tanuj-shardeum:
… malicious validation

### DIFF
--- a/src/utils/functions/signs.ts
+++ b/src/utils/functions/signs.ts
@@ -1,0 +1,15 @@
+import { Sign } from '../../state-manager/state-manager-types'
+
+export function removeDuplicateSignatures(signatures: Sign[]): Sign[] {
+  const seenOwners = new Set<string>()
+  const uniqueSignatures = []
+
+  for (const sign of signatures) {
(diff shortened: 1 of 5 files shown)


src/utils/functions/signs.ts
```diff
@@ -0,0 +1,17 @@
+import { Sign } from '../../state-manager/state-manager-types'
+
+export function removeDuplicateSignatures(signatures: Sign[]): Sign[] {
+  const seen = new Set<string>()
+  const uniqueSignatures = []
+
+  for (const sign of signatures) {
+    const key = `${sign.owner}-${sign.sig}`
```
arhamj: Checking the presence of the owner in the set should be good enough. As one signer can only produce one valid payload for a given payload.

tanuj-shardeum: Done

aniketdivekar (dismissed):

S0naliThakur (dismissed):

S0naliThakur (approved):

PR #317, opened 2024-10-28
SHARD-900: Poqo fixes/improvements

Glitch18:
Fixes and changes to facilitate further investigation

### DIFF
--- a/src/state-manager/TransactionConsensus.ts
+++ b/src/state-manager/TransactionConsensus.ts
@@ -3619,7 +3619,6 @@ class TransactionConsenus {
 
       // save our vote to our queueEntry
       this.crypto.sign(ourVote)
-      this.crypto.sign(proposal)
       queueEntry.ourVote = ourVote
       queueEntry.ourProposal = proposal
       if (queueEntry.firstVoteReceivedTimestamp === 0) {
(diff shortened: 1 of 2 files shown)


aniketdivekar (approved):

arhamj (approved):

src/state-manager/TransactionConsensus.ts
```diff
@@ -3751,60 +3758,60 @@ class TransactionConsenus {
...
-      }
-      return this.crypto.hash(voteToHash)
-    } else {
-      const appliedVote = vote as AppliedVote
-      const voteToHash = Object.assign({}, appliedVote)
-      if (voteToHash.node_id != null) voteToHash.node_id = ''
-      if (voteToHash.sign != null) delete voteToHash.sign
-      return this.crypto.hash(voteToHash)
+  calculateVoteHash(vote: Proposal): string {
+    // if (this.stateManager.transactionQueue.usePOQo && (vote as Proposal).applied !== undefined) {
```
kun6fup4nd4, line 3762: why this condition (vote as Proposal).applied !== undefined) was removed ?

Glitch18: That condition was in place to differentiate between vote being of the Proposal or AppliedVote type, when the function supported both those types. Now that AppliedVote is deprecated, that function was modified to only support Proposal. So that condition to check if vote is of type Proposal is no longer needed.

PR #320, opened 2024-11-07
SHARD-951: Push validator memory info to monitor

shawnifill:
The purpose of this work is to display the validator node's memory information on the monitor node load screen. It will provide additional information that can be used when debugging transaction processing.

### DIFF
--- a/src/reporter/index.ts
+++ b/src/reporter/index.ts
@@ -203,7 +203,7 @@ class Reporter {
       try {
         this.mainLogger.debug(JSON.stringify({report: JSON.stringify(report), method: `_sendReport`, script: '_sendReport'}))
       } catch (error){
-        console.error(error)
+        console.error(`[reporter/index][_sendReport] ERROR while writing report object to log.`, error)
       }
     }
 
--- a/src/reporter/index.ts
+++ b/src/reporter/index.ts
@@ -23,6 +23,7 @@ import { getSocketReport } from '../utils/debugUtils'
 import { Utils } from '@shardus/types'
 import { finishedSyncingCycle } from '../p2p/Join'
 import { currentCycle } from '../p2p/CycleCreator'
+import process from 'process'
 
 const http = require('../http')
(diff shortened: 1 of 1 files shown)


src/reporter/index.ts
```diff
@@ -187,15 +188,25 @@ class Reporter {
...
     }
     const nodeId = Self.id
+    if (logFlags.debug) this.mainLogger.debug(JSON.stringify({method: '_sendReport', script: 'reporter/index', nodeId: nodeId}))
     if (!nodeId) throw new Error('No node ID available to the Reporter module.')
     const report = {
       nodeId,
       data,
     }
+    if (logFlags.debug){
+      try {
```
aniketdivekar, line 203: any particular reason this log is in try catch block? 
if its needed, can we add some prefix text in console.error line in catch block?

shawnifill: This logging is in a try/catch because I did not want an exception from JSON.stringify to cause an issue with executing the remainder of the code. 

I'm logging this way as JSON because Zach and I have discussed improving logging for ingestion into some logging service. Right now, the way we log makes it difficult to create metrics and parse logs in services like AWS Cloudwatch and Google Cloud Monitoring. Writing logs out in a JSON format will make this easier, as will searching for logs specific to a function or script.

shawnifill: @aniketdivekar is this a good enough explanation?

aniketdivekar: Yes, can we add some text prefix to [this](https://github.com/shardeum/shardus-core/pull/320/files#diff-3c0b343c8b320621fa514bd5ee3e66ae28f6f5c68facd9e933f819628672f139R206) error message as identifier
we should be good to go after that

shawnifill: I pushed a change to update the console.error line to have more info.

aniketdivekar (approved):

urnotsam (approved):

PR #322, opened 2024-11-08
SHARD=1088: Check signedReceipt is null before trying to access its properties

urnotsam:
… globalModification


mhanson-github (approved):
Looks good.

PR #329, opened 2024-11-14
SHARD-1068 - get rid of Math.random

kun6fup4nd4:


src/p2p/Archivers.ts
```diff
@@ -85,7 +85,8 @@ export function getArchiverWithPublicKey(
 export function getRandomArchiver(): P2P.ArchiversTypes.JoinedArchiver | null {
   if (archivers.size === 0) return null
   const list = Array.from(archivers.values())
-  return list[Math.floor(Math.random() * list.length)]
+  const index = randomInt(0, list.length);
```
devendra-shardeum, line 88: random no. will include the maximum upper bound.. so it should be list.length -1 .. otherwise.. if the random number is exactly list.length then list[index] will throw error

devendra-shardeum (approved):
done.. looks good

mhanson-github (approved):

urnotsam (approved):

PR #349, opened 2024-12-10
SHARD-1137: Fix getTxTimestampBinary bucket size

mgthuramoemyint:


src/state-manager/TransactionConsensus.ts
```diff
@@ -1811,7 +1811,20 @@ class TransactionConsenus {
       nestedCountersInstance.countEvent('consensus', 'get_tx_timestamp found tx timestamp in cacheById')
       return tsReceipt
     }
-
+    if (Math.abs(cycleCounter - CycleChain.newest.counter) > 1 ){
```
aniketdivekar, line 1815: can we please add some comments around logic to understand what's happening?

src/config/server.ts
```diff
@@ -158,7 +158,8 @@ const SERVER_CONFIG: StrictServerConfiguration = {
     timestampCacheFix: true,
     useAjvCycleRecordValidation: true,
     networkTransactionsToProcessPerCycle: 20,
-    getTxTimestampTimeoutOffset: 0
+    getTxTimestampTimeoutOffset: 0,
+    timestampCacheFixsize: 10000
```
urnotsam: can you fix the casing of this.  Size should have an uppercase 'S'

jintukumardas:
@mgthuramoemyint  Phillip mentioned about adding the rate limit  in the ticket: `rate limit to no more than x requests per second per node` , however, this is not included here. Are we planning to create a separate PR for this?

src/state-manager/TransactionConsensus.ts
mgthuramoemyint: Do you mean in the code or here?

src/config/server.ts
mgthuramoemyint: I have fixed the following issue.

src/state-manager/TransactionConsensus.ts
mgthuramoemyint: Allowing some tx from cycle counter +1 and -1 range.

mgthuramoemyint:
> @mgthuramoemyint Phillip mentioned about adding the rate limit in the ticket: `rate limit to no more than x requests per second per node` , however, this is not included here. Are we planning to create a separate PR for this?

The rate-limit in binary handler might need new design, out of scope for this task.

src/state-manager/TransactionConsensus.ts
aniketdivekar: yeah, i meant in the code

urnotsam (dismissed):

aniketdivekar (dismissed):

PR #351, opened 2024-12-13
Remove refute v2

BelfordZ:


src/p2p/NodeList.ts
```diff
@@ -305,13 +305,33 @@ export function removeNodes(
...
 export function updateNode(
   update: P2P.NodeListTypes.Update,
   raiseEvents: boolean,
   cycle: P2P.CycleCreatorTypes.CycleRecord | null
 ) {
-  const node = nodes.get(update.id)
+  const node = nodes.get(update.id) as NodeWithRefuteCycles
   if (node) {
+    // Initialize refuteCycles if it doesn't exist
+    if (!node.refuteCycles) {
```
afostr, line 319: I think this should be in the if statement below as it can impact the hash of a node

afostr: or at least inside of config.p2p.enableProblematicNodeRemoval

PR #352, opened 2024-12-14
Remove refute v3

BelfordZ:


src/config/server.ts
```diff
@@ -71,6 +71,13 @@ const SERVER_CONFIG: StrictServerConfiguration = {
     maxRotatedPerCycle: 1,
     flexibleRotationDelta: 1,
     flexibleRotationEnabled: false,
+    enableProblematicNodeRemoval: true,
```
afostr: please set this to false for what we merge, then we will either put this in a migration or just set it dynamically after 100% of nodes are 1.16.2

src/p2p/NodeList.ts
```diff
@@ -307,13 +307,34 @@ export function removeNodes(
...
+// create shorthand type for node with refuteCycles
+export type NodeWithRefuteCycles = P2P.NodeListTypes.Node
+
 export function updateNode(
   update: P2P.NodeListTypes.Update,
   raiseEvents: boolean,
   cycle: P2P.CycleCreatorTypes.CycleRecord | null
 ) {
-  const node = nodes.get(update.id)
+  const node = nodes.get(update.id) as NodeWithRefuteCycles
```
afostr: extremely minor, but this can go away now that you have updates lib-types

src/p2p/Rotation.ts
```diff
@@ -81,14 +81,28 @@ export function updateRecord(
...
+    const { expired, removed } = getExpiredRemovedV2(prev, lastLoggedCycle, txs, info)
+    nestedCountersInstance.countEvent('p2p', `results of getExpiredRemovedV2: expired: ${expired} removed: ${removed.length}`, 1)
+    if (logFlags && logFlags.verbose) console.log(`results of getExpiredRemovedV2: expired: ${expired} removed: ${removed.length} array: ${removed}`)
+      
+    record.expired = expired
+    record.removed = removed // already sorted
+  } else {
+    const { expired, removed, problematic } = getExpiredRemovedV3(prev, lastLoggedCycle, txs, info)
+    nestedCountersInstance.countEvent('p2p', `results of getExpiredRemovedV2: expired: ${expired} removed: ${removed.length} problematic: ${problematic}`, 1)
+    if (logFlags && logFlags.verbose) console.log(`results of getExpiredRemovedV2: expired: ${expired} removed: ${removed.length} array: ${removed} problematic: ${problematic}`)
```
afostr: if  you replace:
if (logFlags && logFlags.verbose) console.log(  <the message>
with:
/* prettier-ignore */ if(logFlags?.node_rotation_debug) logger.mainLog_debug('UNIQUE_TAG_HERE', <the message>

it will make evaluating this and other rotation related systems much easier on busy production nodes.
enabled with:
http://localhost:9001/log-flag?name=node_rotation_debug&value=true
(or by the ui in the debug app.. I still need to merge this flag to debug app dev)

afostr: on a large, live, busy network the spam of turning on verbose is quite high.   The above helped me validate all my recent work to rotation etc.

afostr: also this will mean the context of this system will be linked to my other recent logs

src/config/server.ts
BelfordZ: sounds good. Ill put the onCycle to 20k too just so when its turned on, it will still have the delayed fuse.

PR #353, opened 2024-12-16
SHARD-1128: support removing NGTs thru single node's debug endpoint by using gossip

ahmxdiqbal:


src/p2p/ServiceQueue.ts
```diff
@@ -432,6 +507,28 @@ export function sendRequests(): void {
...
   removeProposals.length = 0
+
+  for (const dropNGTInfo of debugDropNGTs) {
+    const unsignedRemoveNetworkTx = {
+      txHash: dropNGTInfo.txHash,
+      cycle: dropNGTInfo.usePrevCycle ? CycleChain.newest.counter - 1 : CycleChain.newest.counter,
+    }
+    txRemove.push(unsignedRemoveNetworkTx)
+
+    Comms.sendGossip(
```
aniketdivekar, line 518: is it possible that `Comms.sendGossip` throws error? if so, shouldn't it be caught & logged? with this implementation, what happens in error case? Do we ignore all the NGTs to be dropped? 

same question goes [here](https://github.com/shardeum/shardus-core/pull/353/files#diff-0fd11e27abb2fd84b758cf52d189a7661521d10f3dd2c035771a04223fceecadR226) also

src/p2p/ServiceQueue.ts
```diff
@@ -727,6 +824,78 @@ function sortedInsert(
   }
 }
 
+function verifyDebugDropNGT(reqParamsDropNGT, cycle): number {
```
urnotsam: It would be better if this could return a object like `{'success': true/false, reason: ''}` instead of arbitrary numerical response.  it will simplify the usage of this function to not have to check each response (0,1,2) when you can instead check if (response) and log out reason if needed.

src/p2p/ServiceQueue.ts
ahmxdiqbal: Comms.sendGossip has a try catch inside it that already handles errors

src/p2p/ServiceQueue.ts
ahmxdiqbal: good point. pushed changes

ahmxdiqbal:
reminder for myself: need to make this rotation safe

src/p2p/ServiceQueue.ts
```diff
@@ -179,13 +183,64 @@ const removeTxGossipRoute: P2P.P2PTypes.GossipHandler<P2P.ServiceQueueTypes.Sign
   }
 }
 
+const debugDropNGTGossipRoute: P2P.P2PTypes.GossipHandler<any> = async (payload, sender, tracker) => {
+  profilerInstance.scopedProfileSectionStart('serviceQueue - debugDropNGT')
+
+  if ([1, 2].includes(currentQuarter) === false) {
+    /* prettier-ignore */ if (logFlags.error) info('debug-drop-network-tx: Got request after quarter 2')
```
aniketdivekar: with `logFlags.error` condition, we are using `info` to log message. is that fine?

src/p2p/ServiceQueue.ts
```diff
@@ -179,13 +183,64 @@ const removeTxGossipRoute: P2P.P2PTypes.GossipHandler<P2P.ServiceQueueTypes.Sign
...
+    return
+  }
+  if (txRemove.some((entry) => entry.txHash === payload.txHash)) {
+    /* prettier-ignore */ if (logFlags.error) info('debug-drop-network-tx: txHash already exists in txRemove')
+    return
+  }
+  if (payload.txHash == null) {
```
aniketdivekar: this check can be early check ( i mean we can move it to top)

src/p2p/ServiceQueue.ts
```diff
@@ -179,13 +183,64 @@ const removeTxGossipRoute: P2P.P2PTypes.GossipHandler<P2P.ServiceQueueTypes.Sign
...
+    /* prettier-ignore */ if (logFlags.error) info('debug-drop-network-tx: txHash not found')
+    return
+  }
+  let cycle = CycleChain.newest
+  if (payload.usePrevCycle) {
+    cycle = CycleChain.getCycleChain(cycle.counter - 1, cycle.counter - 1)[0]
+  }
+  const verificationResult = verifyDebugDropNGT(payload, cycle)
+  if (verificationResult.success === false) {
+    console.log(`debug-drop-ngt - ${verificationResult.message}`)
```
aniketdivekar: should we move this behind logging flag?

src/p2p/ServiceQueue.ts
```diff
@@ -727,6 +823,96 @@ function sortedInsert(
...
+    count: reqParamsDropNGT.sigCounter,
+    nodes: reqParamsDropNGT.pubKeys,
+    networkId: cycle.networkId,
+    cycleCounter: cycle.counter,
+  }
+  const hash = crypto.hash(Utils.safeStringify(payload))
+  const devPublicKeys = getDevPublicKeys() // This should return list of public keys
+  const requestSig = reqParamsDropNGT.sig
+
+  if (reqParamsDropNGT.owner !== '') {
```
aniketdivekar, line 839: looks like logic in this condition and loop below is exactly the same... if so, could you move this in a method instead of duplicating the logic?

src/p2p/ServiceQueue.ts
ahmxdiqbal: imo its fine in this case, this is just receiving gossip a bit late, not some real bug

src/p2p/ServiceQueue.ts
ahmxdiqbal: good point. ill do that next time I get on, its very late(technically really early)

src/p2p/ServiceQueue.ts
ahmxdiqbal: yeah i guess, but we probably want to make sure that the flag we put it behind is usually set to true. this debug drop NGT endpoint isn't a normal flow, its an exceptional one we need to trigger, so if something goes wrong(especially in a prod network), we want to be notified in pretty much every situation. speaking of which, this should also have a counter

src/p2p/ServiceQueue.ts
ahmxdiqbal: ok

src/p2p/ServiceQueue.ts
ahmxdiqbal: now that I think about it, we have a warn log that would probably be best here

src/p2p/ServiceQueue.ts
ahmxdiqbal: after looking at the code closer, I remembered that theyre not actually exactly the same. There's a small difference. if you want, I can refactor it to make it more compact and avoid code duplication which wont be hard, but it is not as simple as putting it in a separate function

ahmxdiqbal: I also dont like the code duplication, but because of the way the logic needed to be implemented, it was the simplest way

urnotsam (approved):

aniketdivekar (approved):

PR #360, opened 2025-01-03
SHARD-1192: Remove deprecated and vulnerable endpoint

Glitch18:
https://linear.app/shm/issue/SHARD-1192/add-accountid-verification-to-the

### DIFF
--- a/src/state-manager/TransactionQueue.ts
+++ b/src/state-manager/TransactionQueue.ts
@@ -515,88 +515,88 @@ class TransactionQueue {
 
     this.p2p.registerInternalBinary(broadcastStateRoute.name, broadcastStateRoute.handler)
 
-    const broadcastFinalStateRoute: P2PTypes.P2PTypes.Route<InternalBinaryHandler<Buffer>> = {
-      name: InternalRouteEnum.binary_broadcast_finalstate,
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      handler: (payload, response, header, sign) => {
-        const route = InternalRouteEnum.binary_broadcast_finalstate
-        nestedCountersInstance.countEvent('internal', route)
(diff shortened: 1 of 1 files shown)


abdulazeem-tk4vr (approved):

src/state-manager/TransactionQueue.ts
```diff
@@ -6754,9 +6754,10 @@ class TransactionQueue {
                     if (configContext.stateManager.attachDataToReceipt === false) {
                       if (configContext.p2p.useFactCorrespondingTell) {
                         this.factTellCorrespondingNodesFinalData(queueEntry)
-                      } else {
-                        this.tellCorrespondingNodesFinalData(queueEntry)
-                      }
+                      } 
+                      // else {
```
aniketdivekar, line 6758: why is this commented?

src/state-manager/TransactionQueue.ts
```diff
@@ -5146,19 +5146,19 @@ class TransactionQueue {
             }
 
 
-          if (this.usePOQo) {
+          // if (this.usePOQo) {
```
aniketdivekar, line 5149: is this a default setting now? can we confirm this with Andrew/Zach before removing this condition?

src/state-manager/TransactionQueue.ts
Glitch18: It's deprecated. We use FACT algorithm and the older function is commented out.

src/state-manager/TransactionQueue.ts
Glitch18: Yes, this is the default setting. We have been removing non POQo branches of code.

aniketdivekar (approved):

urnotsam (approved):

PR #362, opened 2025-01-03
SHARD-1166: Add archiver whitelisting

jintukumardas:


src/p2p/Archivers.ts
```diff
@@ -58,6 +61,7 @@ export let connectedSockets = {}
 let lastSentCycle = -1
 let lastTimeForwardedArchivers = []
 export const RECEIPT_FORWARD_INTERVAL_MS = 5000
+const ALLOWED_ARCHIVERS_UPDATE_INTERVAL_MS = 10000
```
jintukumardas: we can change this to every minute

src/p2p/Archivers.ts
```diff
@@ -269,6 +361,12 @@ export function addArchiverJoinRequest(joinRequest: P2P.ArchiversTypes.Request,
     }
   }
 
+  const isPublicKeyWhitelisted = allowedArchivers.some(archiver => archiver.publicKey === joinRequest.nodeInfo.publicKey);
+  if (!isPublicKeyWhitelisted && archivers.size > 0) {
```
aniketdivekar, line 390: i think we shouldn't have `archivers.size > 0` condition.  because irrespective of length of the archivers, if public key of archiver is not found, it means its invalid.
If archiver length itself is 0 that means something is wrong with the network or our periodic job isn't working (which shouldn't be the case)

jintukumardas: This condition is required to allow the genesis archiver join the network.

aniketdivekar (dismissed):

src/p2p/Archivers.ts
```diff
@@ -33,8 +33,9 @@ import { reportLostArchiver } from '../p2p/LostArchivers/functions'
 import { ActiveNode } from '@shardeum-foundation/lib-types/build/src/p2p/SyncTypes'
 import { Result, ResultAsync } from 'neverthrow'
 import { Utils } from '@shardeum-foundation/lib-types'
-import { arch } from 'os'
 import { checkGossipPayload } from '../utils/GossipValidation'
+import { arch } from 'os'
```
urnotsam: is this being used?

jintukumardas: Removed

urnotsam (approved):

src/p2p/Archivers.ts
```diff
@@ -126,6 +147,94 @@ export function init() {
...
+  }
+
+  // Helper function to fetch and validate archiver data
+  async function fetchArchiverData(ip: string, port: number): Promise<ArchiverResponse | null> {
+    try {
+      const response = await fetch(`http://${ip}:${port}/allowed-archivers`)
+      if (!response.ok) {
+        return null
+      }
+      const data = await response.json() as ArchiverResponse
```
mssabr01, line 201: This needs a check on the length of the body to prevent large bodies from eating up memory and cycles. Additionally, please write an AJV definition for ArchiverResponse and use that to verify this response

src/p2p/Archivers.ts
```diff
@@ -169,6 +278,14 @@ export function updateRecord(txs: P2P.ArchiversTypes.Txs, record: P2P.CycleCreat
     .filter((request) => request.requestType === P2P.ArchiversTypes.RequestTypes.LEAVE)
     .map((leaveRequest) => leaveRequest.nodeInfo)
 
+  // Remove archivers that are not in the allowed list
+  for (const existingArchiver of archivers.values()) {
+    if (allowedArchivers.length > 0 && !allowedArchivers.some(allowed => allowed.publicKey === existingArchiver.publicKey) &&
```
mssabr01, line 283: If allowedArchivers length is 0 then this fails and nothing changes. I believe if allowedArchivers list is empty then that should mean no archivers are allowed and this node has failed its sync/should leave network.

src/p2p/Archivers.ts
mssabr01: The genesis archiver would still be in the whitelist. This condition allows the genesis archiver to be anyone. Its not a likely scenario but still can happen. How would removing this condition prevent the genesis archiver from joining?

src/p2p/Archivers.ts
jintukumardas: This condition is added to allow the genesis archiver to join

jintukumardas: This condition was added so that nodes don't add the genesis archiver to the leaving archiver list since the archiver list is empty during network startup. 

Also, this update is propagated through cycleRecord. If the node is OOS, this cycle record will not be accepted.

jintukumardas: The node fetches the allowed archiver list directly from the archiver (randomly chosen), the list shouldn't be empty unless the signatures are invalid.

PR #365, opened 2025-01-07
Fix/problem node tracker

BelfordZ:
# Problem Node Tracker Bug Fixes

### Problem
- `updateNode` was not being called every cycle for every node
- Only called when there are changes to node properties
- Critical issue: nodes being marked as refuted/lost did not trigger an `updateNode` call
- This led to incomplete tracking of problematic nodes

### Solution
- Added `updateProblematicNodeTracking` in two strategic locations:
  1. Inside [`Sync::applyNodeListChange`](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/Sync.ts#L486-L491)
  2. During node updates via [`NodeList.updateProblematicNodeTracking`](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/NodeList.ts#L379-L405)
- This ensures consistent tracking regardless of node state changes
- Key implementation details:
  - [Tracking function implementation](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/NodeList.ts#L379-L405)
  - [Integration with Sync module](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/Sync.ts#L486-L491)

### Problem
- `refuteCycles` initialization in `updateNode` was problematic
- Nodes could exist without initialized `refuteCycles`
- Activation cycle check was causing initialization delays

### Solution
- Moved initialization to [`addNode` function](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/NodeList.ts#L115-L119)
- Removed activation cycle check
- Initialize immediately if problematic node tracking is enabled
- Ensures consistent state from node creation
- Key changes:
  - [Initialization code](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/NodeList.ts#L115-L119)
  - [Configuration flag check](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/config/server.ts#L71-L77)

### Problem
- NodeList gossip mechanism broken due to `Set` serialization
- `refuteCycles` as `Set` couldn't be properly transmitted

### Solution
- Changed `refuteCycles` from `Set` to `Array`
- Ensures proper serialization during gossip
- Maintains chronological order of refute cycles
- Improves rotation safety
- Implementation details:
  - [Array conversion check](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/NodeList.ts#L386-L388)
  - [Updated problematic node handling](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/ProblemNodeHandler.ts#L6-L24)
  - [Array-based refute tracking](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/p2p/NodeList.ts#L394-L397)

## Implementation Notes
- These changes are interdependent and should be deployed together
- The switch from `Set` to `Array` requires careful handling of duplicates
- Performance impact should be minimal as array operations are still O(n)
- Migration path includes automatic conversion from `Set` to `Array` for existing nodes
- Key configuration:
  - [Server configuration updates](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/config/server.ts#L71-L77)
  - [Type definitions](https://github.com/shardeum/shardus-core/blob/bf4c075419ea15f7905c4c7149e40b1351c9f40d/src/shardus/shardus-types.ts#L759-L766)

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "@mapbox/node-pre-gyp": "1.0.10",
     "@shardus/crypto-utils": "git+https://github.com/shardeum/lib-crypto-utils#itn4",
     "@shardus/net": "git+https://github.com/shardeum/lib-net#itn4",
-    "@shardus/types": "git+https://github.com/shardeum/lib-types#itn4",
+    "@shardus/types": "git+https://github.com/shardeum/lib-types#itn4-1.16.3",
     "@types/better-sqlite3": "7.6.3",
     "body-parser": "1.18.3",
     "bytenode": "1.3.4",
@@ -109,6 +109,9 @@
     "webpack-node-externals": "1.7.2"
   },
   "overrides": {
+    "@shardus/crypto-utils": {
+      "@shardus/types": "git+https://github.com/shardeum/lib-types#itn4-1.16.3"
+    },
     "qs": "6.11.2",
     "ip": "2.0.1",
     "@hapi/hoek": "8.5.1",
--- a/src/config/server.ts
+++ b/src/config/server.ts
@@ -71,6 +71,7 @@ const SERVER_CONFIG: StrictServerConfiguration = {
     maxRotatedPerCycle: 1,
     flexibleRotationDelta: 1,
     flexibleRotationEnabled: false,
+    enableDangerousProblematicNodeRemoval: false,
     enableProblematicNodeRemoval: false,
     enableProblematicNodeRemovalOnCycle: 20000,
     maxProblematicNodeRemovalsPerCycle: 1,
--- a/src/p2p/ModeSystemFuncs.ts
+++ b/src/p2p/ModeSystemFuncs.ts
@@ -594,16 +594,25 @@ export function getExpiredRemovedV2(
   return { expired, removed }
 }
 
-/** Returns the number of expired nodes and the list of removed nodes using calculateToAcceptV2 
- *  this list includes problematic nodes + expired nodes.
+type GetExpiredRemovedV3Result = {
+  // the number of problematic nodes that we have removed
+  problematic: number
+  // the number of expired nodes that we have removed
+  expired: number
+  // the list of nodes (by id) that we have removed
+  removed: string[]
+}
+
+/** 
+ * Gets the list of nodes that should be removed from the network in the next cycle.
+ * 
 */
 export function getExpiredRemovedV3(
   prevRecord: P2P.CycleCreatorTypes.CycleRecord,
   lastLoggedCycle: number,
   txs: P2P.RotationTypes.Txs & P2P.ApoptosisTypes.Txs,
   info: (...msg: string[]) => void
-): { problematic: number; expired: number; removed: string[] } {
-
+): GetExpiredRemovedV3Result {
   // clear state from last run
   NodeList.potentiallyRemoved.clear()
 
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "@mapbox/node-pre-gyp": "1.0.10",
     "@shardus/crypto-utils": "git+https://github.com/shardeum/lib-crypto-utils#itn4",
     "@shardus/net": "git+https://github.com/shardeum/lib-net#itn4",
-    "@shardus/types": "git+https://github.com/shardeum/lib-types#itn4",
+    "@shardus/types": "git+https://github.com/shardeum/lib-types#e7b0063490d2885d2f1eaa43d389c5f5dfe6c0fb",
```
BelfordZ: requires https://github.com/shardeum/lib-types/pull/26

src/p2p/ModeSystemFuncs.ts
```diff
@@ -670,11 +674,32 @@ export function getExpiredRemovedV3(
...
+    numProblematicRemovals: numProblematicRemovals,
+    numProblematicNodes: problematicNodes.length,
+    canRemoveProblematicNodesThisCycle: canRemoveProblematicNodesThisCycle,
+    numPotentiallyExpiredRemovals: numPotentiallyExpiredRemovals,
+    numExpiredRemovals: numExpiredRemovals,
+  }
   nestedCountersInstance.countEvent(
-      'p2p',
-      `results of getExpiredRemovedV3: numApoptosizedRemovals: ${numApoptosizedRemovals}, numProblematicRemovals: ${numProblematicRemovals}, numExpiredRemovals: ${numExpiredRemovals}, removed: ${remove}`
+    'p2p',
```
BelfordZ: I dont think there will be too many permutations of this during a nodes 'tour of duty'. .. but we could put this behind a flag too if desireable... but its very useful debugging data.

src/p2p/ModeSystemFuncs.ts
```diff
@@ -698,5 +723,5 @@ export function getExpiredRemovedV3(
     insertSorted(removed, node.id)
   }
 
-  return { problematic: problematicNodes.length, expired: numExpiredNodes, removed }
+  return { problematic: numProblematicRemovals, expired: numExpiredRemovals, removed }
```
BelfordZ, line 728: note: we are reporting the number of node we removed as a result of being expired or problematic. 

As it was, it was returning the number of expired nodes in total (which is very often much less than how many we actually remove). I believe this is what was thowing off the charting.. but I could be wrong there.

src/config/server.ts
```diff
@@ -71,6 +71,7 @@ const SERVER_CONFIG: StrictServerConfiguration = {
     maxRotatedPerCycle: 1,
     flexibleRotationDelta: 1,
     flexibleRotationEnabled: false,
+    enableDangerousProblematicNodeRemoval: false,
     enableProblematicNodeRemoval: false,
     enableProblematicNodeRemovalOnCycle: 20000,
```
BelfordZ, line 76: Should set this higher just to prevent any accidents

BelfordZ: to infinity!

src/p2p/ModeSystemFuncs.ts
```diff
@@ -670,11 +674,32 @@ export function getExpiredRemovedV3(
...
+    numExpiredRemovals: numExpiredRemovals,
+  }
   nestedCountersInstance.countEvent(
-      'p2p',
-      `results of getExpiredRemovedV3: numApoptosizedRemovals: ${numApoptosizedRemovals}, numProblematicRemovals: ${numProblematicRemovals}, numExpiredRemovals: ${numExpiredRemovals}, removed: ${remove}`
+    'p2p',
+    `results of getExpiredRemovedV3: ${JSON.stringify(counters)}`
   )
 
+  const dangerousRemoval = remove === 0 && config.p2p.enableDangerousProblematicNodeRemoval === false
```
BelfordZ, line 694: this variable name is confusing

It should really be 'isSafeRemoval'

why? Because what we are saying here is, unless this dangerousRemove FF is ON, we will return early if its suggested not to remove any nodes.

PR #378, opened 2025-02-04
SHARD-1128: remove usePreCycle usage

ahmxdiqbal:


src/p2p/ServiceQueue.ts
```diff
@@ -894,9 +900,18 @@ function verifyDebugDropNGT(reqParamsDropNGT, cycle): { success: boolean; messag
...
       sign,
-    } as SignedObject
+    }
 
-    const verified = crypto.verify(hashIncluded, hashIncluded.sign.owner)
+    let verified = crypto.verify(hashIncluded, hashIncluded.sign.owner)
+    if (verified === false) {
+      delete hashIncluded.requestHash
+      delete hashIncluded.sign
+      hashIncluded.cycleCounter--
```
aniketdivekar, line 911: any references to why we are doing this? (decrementing the counter)
it would help if you write some comment on what exactly are we doing when verified is false with `hashIncluded.cycleCounter`

ahmxdiqbal: Basically, the `hashIncluded` object contains the cycle number. We only get the signature from the debug app, not the original object, so we need to reconstruct what the object would have been. One of the field of `hashIncluded` is the cycle counter. the trouble is that there are two possible values for the cycle counter based upon whether the node that received the drop ngt request gossiped it within the cycle it received it or waited for the next cycle. So since there are two possible values, we need to verify the signature with both of them. So if the case that it fail verification with the current cycle counter, we check with the previous cycle counter

aniketdivekar (approved):

urnotsam (approved):

PR #385, opened 2025-02-11
SHARD-1866: Prevent stuck NGTs in consensing state

urnotsam:


PudgyPug (dismissed):

src/state-manager/TransactionQueue.ts
```diff
@@ -6845,6 +6849,17 @@ class TransactionQueue {
...
                   //just keep waiting for a reciept
+                  if (queueEntry.isNGT && txAge > timeM * 5) {
+                    // entry is an NGT so we want to remove it if consensing fails to prevent from getting stuck
+                    nestedCountersInstance.countEvent(
+                      `consensus`,
+                      'removing NGT from queue after failed consensing'
+                    )
+                    this.updateTxState(queueEntry, 'fail')
+                    this.removeFromQueue(queueEntry, currentIndex)
+                    this.processQueue_clearAccountsSeen(seenAccounts, queueEntry)
```
afostr, line 6857: new functionality should be behind a config.    please also use a new config for the timeout time (replace 5)

src/state-manager/TransactionQueue.ts
```diff
@@ -1943,6 +1943,9 @@ class TransactionQueue {
         }
       }
 
+      const INIT_REWARD_TX = 8
+      const CLAIM_REWARD_TX = 9
+      const NGT_TYPES = [INIT_REWARD_TX, CLAIM_REWARD_TX]
```
afostr: do we have any way of knowing if the TX is NGT.   this breaks the separation of biz logic and the core protocol.    and app defined isNGT(tx):bool. function would solve this, but open to other ways.

urnotsam: currently, no.  Te only way to identify it is by knowing out internal tx types that we submit by NGT.  We could submit the transaction (.put(tx)) with a field that identifies it as an NGT, but thats really no different than this approach when you boil it down.

src/state-manager/TransactionQueue.ts
```diff
@@ -2047,7 +2050,8 @@ class TransactionQueue {
         hasRobustConfirmation: false,
         sharedCompleteData: false,
         correspondingGlobalOffset: 0,
-        isSenderWrappedTxGroup: {}
+        isSenderWrappedTxGroup: {},
+        isNGT: NGT_TYPES.includes(acceptedTx.data?.tx?.['internalTXType']),
```
afostr: this is the part than can be replaced with an app.isNGT function

urnotsam: okay

afostr (dismissed):
looks great

afostr (approved):
looks good

PudgyPug (approved):

PR #386, opened 2025-02-11
SHARD-1868: add bad node perf csv logging v1

ahmxdiqbal:


src/logger/csvPerfEvents.ts
```diff
@@ -0,0 +1,27 @@
...
+import { config } from './../p2p/Context'
+import { currentCycle } from '../p2p/CycleCreator'
+import { logDir_global } from '../shardus'
+
+const csvFilePath = '/nodePerfEvents.csv'
+let firstCall = true
+
+export function logPerfEvents(eventName: string): void {
+  if (config.debug.logCSVPerfEvents) {
+    if (firstCall) {
```
urnotsam: it would probably be better to check if the file exists here and write the header if it doesnt.  In the case where a node is restarted the current approach would write the header again.

src/logger/csvPerfEvents.ts
```diff
@@ -0,0 +1,27 @@
+import * as fs from 'fs'
+import { config } from './../p2p/Context'
+import { currentCycle } from '../p2p/CycleCreator'
+import { logDir_global } from '../shardus'
+
+const csvFilePath = '/nodePerfEvents.csv'
+let firstCall = true
+
+export function logPerfEvents(eventName: string): void {
```
urnotsam, line 10: We should create a buffer that flushes after every X (maybe 10?) perfEvents to prevent too many writes that could create a lot of I/O overhead and DOS the node.  

With this approach we would also want to make sure we flush the buffer when the node shuts down.

urnotsam (changes requested):

src/logger/csvPerfEvents.ts
ahmxdiqbal: yeah this is a better way to handle it, i wasnt thnking about the node restarting

src/logger/csvPerfEvents.ts
ahmxdiqbal: is this really a concern? based upon the perf events we're actually going to use this function for, theyll happen max once a cycle. the only exception is the endpoint, which will be used way less frequently

urnotsam (approved):

mhanson-github (approved):

PR #407, opened 2025-02-24
SHARD-1887: Fix double counting signatures in validateClosestActiveNodeSignatures

jintukumardas:
fix: count signers in validateClosestActiveNodeSignatures instead of signatures


mhanson-github (approved):
Looks gtm.

aniketdivekar (approved):

PR #432, opened 2025-03-17
SHARD-2406 - Add unit tests for shardus-types

S0naliThakur:


BelfordZ (changes requested):
We doint need to test files that only export types. Or at least for now. Lets hit snooze on these.

S0naliThakur:
> We doint need to test files that only export types. Or at least for now. Lets hit snooze on these.

May be we should remove/mark such instances from the list.
cc: @mhanson-github @aniketdivekar

PR #464, opened 2025-04-07
Shard 2195

mssabr01:
Updated binary_sync_trie_hashes to check that the payload it received comes from an active node


mssabr01:
NodeList only has active nodes. Internal routes have active node protections deeper in the protocol

PR #473, opened 2025-04-17
SHARD-2182 : recent restore fixes

afostr:
Resolves issue where nodes were forwarding the wrong best cycle markers when more than one cycle had to be forwarded.
Optional edge case fix to limit recently activated nodes from sending cycle data.


mhanson-github (dismissed):
Reviewed in vid call with andrew.

mhanson-github (approved):
Looks good

